PR #15, opened 2017-01-31
safe null support for include and exclude

pwelter34:


src/Foundatio.Repositories.Elasticsearch/Queries/Builders/FieldIncludesQueryBuilder.cs
```diff
@@ -36,7 +36,9 @@ public class FieldIncludesQueryBuilder : IElasticQueryBuilder {
 
     public static class FieldIncludesQueryExtensions {
         public static T IncludeField<T>(this T query, Field field) where T : IFieldIncludesQuery {
-            query.FieldIncludes?.Add(field);
+            if (field != null)
```
ejsmith: Shouldn't we check for null or empty?

src/Foundatio.Repositories.Elasticsearch/Queries/Builders/FieldIncludesQueryBuilder.cs
```diff
@@ -48,8 +48,11 @@ public static class FieldIncludesQueryExtensions {
         }
 
         public static T IncludeFields<T>(this T query, params Field[] fields) where T : IFieldIncludesQuery {
-            if (fields != null)
-                query.FieldIncludes?.AddRange(fields);
+            if (fields == null)
+                return query;
+
+            foreach (var field in fields.Where(p => !string.IsNullOrEmpty(p.Name)))
```
ejsmith: Use String.IsNullOrEmpty for consistency

src/Foundatio.Repositories.Elasticsearch/Queries/Builders/FieldIncludesQueryBuilder.cs
```diff
@@ -74,8 +77,11 @@ public static class FieldIncludesQueryExtensions {
         }
 
         public static T ExcludeFields<T>(this T query, params Field[] fields) where T : IFieldIncludesQuery {
-            if (fields != null)
-                query.FieldExcludes?.AddRange(fields);
+            if (fields == null)
+                return query;
+
+            foreach (var field in fields.Where(p => !string.IsNullOrEmpty(p.Name)))
```
ejsmith: Use String.IsNullOrEmpty for consistency

pwelter34: is name the correct thing to check?  what if property has a value but not name?

ejsmith: Yeah, you are right, we can't just check Name. Name isn't populated if you use a lambda expression. So not sure what all you would have to check to see if its empty.

PR #19, opened 2017-03-13
ReindexScripts on VersionedIndex

randylsu:


src/Foundatio.Repositories.Elasticsearch/Configuration/VersionedIndex.cs
```diff
@@ -21,6 +22,16 @@ public VersionedIndex(IElasticConfiguration configuration, string name, int vers
         public int Version { get; }
         public string VersionedName { get; }
         public bool DiscardIndexesOnReindex { get; set; } = true;
+        private List<ReindexScript> ReindexScripts { get; } = new List<ReindexScript>();
+    
+        private class ReindexScript {
+            public Func<int, int, bool> ShouldRun { get; set; }
```
ejsmith: This is a bit confusing as it is. Can we at least create a delegate for this so its more intuitive as to what the parameters are?

randylsu:
About to add an overload for Type that wraps the script in an `if (ctx._type == '{type}')` block

niemyjski:
I just added a few types for IPatchOperation I wonder if these could be made more generic and reused?

PR #28, opened 2018-03-12
add a query options to only return ids

derekpitt:


niemyjski:
I thought we had the ability to say here are an included/excluded list of columns to return. Could this be used instead?

derekpitt:
this is a little different..   we have some big documents and even just including one field, elastic search is slowing us down. this skips all of that, so it is way more performant with source: false for our use case

niemyjski:
LGTM

PR #51, opened 2020-02-07
Throw an error if you don't specify a timestamp field which is requir…

niemyjski:
…ed for reindexing.


niemyjski:
@ejsmith I'm merging this because it's better to know earlier that it failed rather than later when the index task randomly fails and doesn't tell you other than all shards failed.

PR #84, opened 2020-06-20
Set object id based on a date func

niemyjski:


src/Foundatio.Repositories.Elasticsearch/Configuration/DailyIndex.cs
```diff
@@ -45,11 +45,11 @@ public DailyIndex(IElasticConfiguration configuration, string name, int version
                 switch (document) {
                     case null:
                         throw new ArgumentNullException(nameof(document));
-                    case IHaveCreatedDate createdDoc when createdDoc.CreatedUtc != DateTime.MinValue:
-                        return createdDoc.CreatedUtc;
                     // This is also called when trying to create the document id.
                     case IIdentity identityDoc when identityDoc.Id != null && ObjectId.TryParse(identityDoc.Id, out var objectId) && objectId.CreationTime != DateTime.MinValue:
                         return objectId.CreationTime;
+                    case IHaveCreatedDate createdDoc when createdDoc.CreatedUtc != DateTime.MinValue:
```
niemyjski, line 51: I moved these around because it aligns with the logic we are doing in other spots (where we have this exact expression)

src/Foundatio.Repositories.Elasticsearch/Configuration/DailyIndex.cs
```diff
@@ -405,6 +405,16 @@ public DailyIndex(IElasticConfiguration configuration, string name, int version
             return GetIndexByDate(date);
         }
 
+        public override string CreateDocumentId(object document) {
+            if (_getDocumentDateUtc != null) {
```
niemyjski, line 409: This should never be null, but we are doing null checks everywhere.

tests/Foundatio.Repositories.Elasticsearch.Tests/QueryableRepositoryTests.cs
```diff
@@ -56,6 +57,23 @@ public sealed class QueryableRepositoryTests : ElasticRepositoryTestBase {
             Assert.Equal(1, await _dailyRepository.CountAsync(q => q.DateRange(utcNow.AddHours(-1), utcNow.AddHours(1), "createdUtc")));
         }
 
+        [Fact]
+        public async Task CanRoundTripById() {
+            using var _ = TestSystemClock.Install();
```
niemyjski, line 62: I was kinda shocked I couldn't just do `using TestSystemClock.Install()`

src/Foundatio.Repositories.Elasticsearch/Configuration/DailyIndex.cs
ejsmith: I guess this is alright. If it has an id it should probably prefer that.

ejsmith (approved):

PR #91, opened 2021-08-12
Cache improvements

ejsmith:
Call AddDocumentsToCacheAsync in more situations. Allow immediately caching added or saved docs. Add docs to cache by Id if they are not coming from dirty reads.

### DIFF
--- a/src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticReadOnlyRepositoryBase.cs
+++ b/src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticReadOnlyRepositoryBase.cs
@@ -71,8 +71,6 @@ namespace Foundatio.Repositories.Elasticsearch {
                 return null;
 
             options = ConfigureOptions(options.As<T>());
-            if (IsCacheEnabled && options.HasCacheKey())
-                throw new ArgumentException("Cache key can't be set when calling GetById");
 
             if (IsCacheEnabled && options.ShouldReadCache()) {
                 var value = await GetCachedFindHit(id).AnyContext();
@@ -100,7 +98,7 @@ namespace Foundatio.Repositories.Elasticsearch {
             }
 
             if (IsCacheEnabled && options.ShouldUseCache())
-                await AddDocumentsToCacheAsync(findHit ?? new FindHit<T>(id, null, 0), options).AnyContext();
+                await AddDocumentsToCacheAsync(findHit ?? new FindHit<T>(id, null, 0), options, false).AnyContext();
 
             return ShouldReturnDocument(findHit?.Document, options) ? findHit?.Document : null;
         }
@@ -118,8 +116,6 @@ namespace Foundatio.Repositories.Elasticsearch {
                 throw new NotSupportedException("Model type must implement IIdentity.");
 
             options = ConfigureOptions(options.As<T>());
-            if (IsCacheEnabled && options.HasCacheKey())
-                throw new ArgumentException("Cache key can't be set when calling GetByIds");
 
             var hits = new List<FindHit<T>>();
             if (IsCacheEnabled && options.ShouldReadCache())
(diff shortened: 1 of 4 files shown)


src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticReadOnlyRepositoryBase.cs
```diff
@@ -426,7 +426,7 @@ public abstract class ElasticReadOnlyRepositoryBase<T> : ISearchableReadOnlyRepo
             result = response.Hits.Select(h => h.ToFindHit()).ToList();
 
             if (IsCacheEnabled && options.ShouldUseCache())
-                await AddDocumentsToCacheAsync(result, options).AnyContext();
+                await AddDocumentsToCacheAsync(result, options, options.GetConsistency(DefaultConsistency) == Consistency.Immediate ? false : true).AnyContext();
```
niemyjski, line 429: Seems kinda odd to me that I'd use a consistency mode when doing a search. Seems like the common behavior would be to cache the results on add/save or pass a consistency mode when calling add/save.

src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticReadOnlyRepositoryBase.cs
```diff
@@ -389,6 +385,10 @@ public abstract class ElasticReadOnlyRepositoryBase<T> : ISearchableReadOnlyRepo
             var nextPageFunc = ((IGetNextPage<TResult>)result).GetNextPageFunc;
             ((IGetNextPage<TResult>)result).GetNextPageFunc = null;
             await SetCachedQueryResultAsync(options, result, cacheSuffix: cacheSuffix).AnyContext();
+            
+            if (IsCacheEnabled && options.ShouldUseCache() && options.GetConsistency(DefaultConsistency) == Consistency.Immediate)
+                await AddDocumentsToCacheAsync((ICollection<T>)result.Documents, options, false);
```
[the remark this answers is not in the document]
ejsmith: Because this query result is already potentially getting cached as a whole result. It seems strange to me to send a potentially large list of documents through the AddDocumentsToCache that are all dirty reads to be added individually to cache. I guess it could be confusing that it doesn't happen, but I think I'd rather argue that documents that go through find and potentially return large results don't go through the AddDocumentsToCache at all even if it's not a dirty read.

src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticReadOnlyRepositoryBase.cs
ejsmith: It's not odd. It's pretty common to say that I want to do something like a FindOne using a query and that I want to use immediate consistency meaning that I will wait for any pending writes.

niemyjski: We've never set `Consistency.Immediate` in any of our apps on a FindOne call to my knowledge.

src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticReadOnlyRepositoryBase.cs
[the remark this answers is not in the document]
ejsmith: This is actually bad because we are already caching the entire result with the specified cache key. Calling this method could interfere with that and also could flood your cache with tons of docs being added by id. I'm going to remove this.

src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticReadOnlyRepositoryBase.cs
ejsmith: During auth we often lookup by email address using a FindOne with immediate consistency.

PR #93, opened 2021-08-31
Adding support for Elasticsearch runtime fields

ejsmith:


src/Foundatio.Repositories.Elasticsearch/Queries/Builders/IElasticQueryBuilder.cs
```diff
@@ -70,17 +74,25 @@ public interface IQueryBuilderContext {
     }
 
     public static class QueryBuilderContextExtensions {
-        public static void SetTimeZone(this IQueryBuilderContext context, string timeZone) {
+        public static void SetTimeZone(this IQueryBuilderContext context, Lazy<string> timeZone) {
```
niemyjski, line 77: We'd need to have an lazy async callback here to get the time zone.

PR #104, opened 2024-05-20
added support for decimal column sorting with SearchAfter & SearchBefore

randylsu:
The ObjectConverter did not support parsing of decimal tokens, which caused DecodeSortToken to throw when sorting on decimal fields while using SearchBefore or SearchAfter paging. This PR adds support for decimals & an integration test case.


ejsmith (approved):
Looks good.

PR #114, opened 2024-12-17
Apply includes and excludes to GetById and GetByIds

ejsmith:


src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticReadOnlyRepositoryBase.cs
```diff
@@ -714,6 +720,79 @@ protected virtual ICommandOptions<T> ConfigureOptions(ICommandOptions<T> options
         return options;
     }
 
+    protected virtual void ConfigureGetRequest(GetRequest request, ICommandOptions options)
+    {
+        var (resolvedIncludes, resolvedExcludes) = GetResolvedIncludesAndExcludes(options);
+
+        if (resolvedIncludes.Length > 0 && resolvedExcludes.Length > 0)
+        {
```
niemyjski, line 728: This seems odd to have both includes and excludes.

niemyjski (commented):
Looks good, do you think we should hash includes and excludes and use it part of the cache key?

src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticReadOnlyRepositoryBase.cs
ejsmith: You can 100% have both.

ejsmith:
> Looks good, do you think we should hash includes and excludes and use it part of the cache key?

Seems like that is over-complicating things and we should do what you asked us to do.

niemyjski (approved):

PR #143, opened 2025-06-18
Added failing index tests to show that indexes can be created without configure indexes being called.

niemyjski:
This took a while to track down and I'm hoping there isn't other cases but have been seeing cases where we are seeing errors like this:

```
"Error creating the index history-v2-2025.06
Original: [ElasticsearchClientException] Request failed to execute. Call: Status code 400 from: PUT /history-v2-2025.06. ServerError: Type: invalid_alias_name_exception Reason: \"Invalid alias name [history-2025.06]: an index or data stream exists with the same name as the alias\"
Server Error (Index=): Invalid alias name [history-2025.06]: an index or data stream exists with the same name as the alias
[400] PUT /history-v2-2025.06`
```

These indexes do script updates.

### DIFF
--- a/tests/Directory.Build.props
+++ b/tests/Directory.Build.props
@@ -8,7 +8,7 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.14.1" />
     <PackageReference Include="xunit" Version="2.9.3" />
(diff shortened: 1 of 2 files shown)


niemyjski (commented):
Self reviewed

PR #161, opened 2025-08-21
Fixed a bug where get by id(s) with cache key would also set the id b…

niemyjski:
…ased cache route when there was an include or exclude defined. Also ensure we look up by the cache key first.


niemyjski (commented):
Self reviewed

PR #244, opened 2026-03-24
Serialize concurrent ConfigureIndexesAsync with distributed lock + cache marker

niemyjski:
## TLDR

When multiple distributed processes (pods, workers, migration runners) call `ConfigureIndexesAsync` on startup, only the first one does real work. Everyone else waits or skips via a distributed lock and configuration-aware cache marker. The marker is cleared on delete and reindex so the next configure call re-validates.

## Problem

`ConfigureIndexesAsync` had no concurrency protection. Every caller ran the full configure pass (index exists checks, create/update settings, put mappings, alias management, reindex detection) regardless of whether another process was already doing the same work. At scale (100 nodes, 100+ indexes), this means thousands of redundant Elasticsearch admin API calls on every deployment, plus racing reindex work item enqueuing.

## Solution

Added a distributed double-checked lock + cache marker pattern to `ElasticConfiguration.ConfigureIndexesAsync`:

1. **Cache check** -- if a configuration marker exists in the distributed cache, return immediately (zero ES calls, zero lock overhead)
2. **Distributed lock** -- acquire via `CacheLockProvider` (1-minute lock duration, 1-minute wait timeout)
3. **Double-check** -- after acquiring the lock, check cache again (another process may have finished while we waited)
4. **Configure** -- run the full `ConfigureAsync` + `MaintainAsync` on all indexes in parallel
5. **Set cache marker** -- 5-minute TTL, only set after successful completion (partial failures retry)

The cache marker key includes a stable `XxHash64` hash of all index names and versions (sorted, built incrementally). Deploying a new configuration (adding indexes, changing versions) automatically produces a different hash, bypassing stale markers from a previous configuration. Old markers expire naturally after 5 minutes.

The marker is explicitly cleared by `DeleteIndexesAsync` and `ReindexAsync` so the next configure call re-runs after any structural change. `MaintainIndexesAsync` does not clear the marker because it does not change index structure.

Cache operations (`ExistsAsync`, `SetAsync`) are wrapped in resilient helper methods (`TryCheckCacheMarkerAsync`, `TrySetCacheMarkerAsync`) that catch non-cancellation exceptions and log warnings, so a cache outage falls back to the pre-lock behavior rather than failing the entire configuration.

### Behavior at scale

| Caller | Cache lookups | Lock ops | ES API calls |
|---|---|---|---|
| Winner (first process) | 2 | acquire + release | Full configure |
| Waiter (during configure) | 2 | acquire + release | 0 |
| Late arrival (after configure) | 1 | 0 | 0 |

### Failure modes

- **Cache down**: Falls back to current behavior (everyone configures). Idempotent, just redundant.
- **Lock timeout (null return)**: Skips configuration, assumes another process is handling it.
- **Partial failure**: Exception propagates, cache marker never set, next caller retries.
- **Manual ES changes**: 5-minute TTL safety net ensures reconfiguration.
- **New deployment**: Different index names/versions produce a new hash, bypassing the old marker automatically.

### Design decisions

- **No lock renewal**: Even 100+ indexes in parallel complete in seconds. 1-minute lock is generous.
- **Explicit `indexes` parameter bypasses lock+cache**: Callers who pass specific indexes are doing targeted work and should not be blocked by or contribute to the "all indexes configured" marker.
- **No interface changes**: `IElasticConfiguration` signature is unchanged.
- **Configuration-aware cache key**: `XxHash64` hash of sorted index names + versions ensures stale markers from previous deployments are automatically bypassed.
- **Separate lock and cache keys**: `ConfigureIndexesLockName` for the distributed lock, `ConfigureIndexesCacheKeyPrefix` + hash for the marker, avoiding `CacheLockProvider` key conflicts.

## Changes

- `ElasticConfiguration.ConfigureIndexesAsync` -- distributed lock + configuration-aware cache marker with double-checked locking
- `ElasticConfiguration.DeleteIndexesAsync` -- clears cache marker (via prefix) after deletion
- `ElasticConfiguration.ReindexAsync` -- clears cache marker (via prefix) after reindexing
- `ElasticConfiguration.GetConfigureIndexesCacheKey` -- incremental `XxHash64` hash of index names + versions
- `ElasticConfiguration.TryCheckCacheMarkerAsync` / `TrySetCacheMarkerAsync` -- resilient cache helpers
- Tests with proper Arrange/Act/Assert verifying marker behavior, skip paths, and invalidation
- Documentation in `index-management.md` and `elasticsearch-setup.md`


src/Foundatio.Repositories.Elasticsearch/Configuration/ElasticConfiguration.cs
```diff
@@ -163,28 +189,32 @@ private async Task ConfigureIndexInternalAsync(IIndex idx, bool beginReindexingO
...
             indexes = Indexes;
 
         var tasks = new List<Task>();
         foreach (var idx in indexes)
             tasks.Add(idx.DeleteAsync());
 
-        return Task.WhenAll(tasks);
+        await Task.WhenAll(tasks).AnyContext();
+
+        await Cache.RemoveAsync(ConfigureIndexesCacheKey).AnyContext();
```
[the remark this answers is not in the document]
niemyjski: Already addressed in a previous commit: \DeleteIndexesAsync\ now uses try/finally to ensure cache marker is cleared even on partial failures.

src/Foundatio.Repositories.Elasticsearch/Configuration/ElasticConfiguration.cs
```diff
@@ -122,16 +123,41 @@ public IIndex GetIndex(string name)
...
+        }
+
+        var allTasks = new List<Task>();
+        foreach (var idx in Indexes)
+            allTasks.Add(ConfigureIndexInternalAsync(idx, beginReindexingOutdated));
+
+        await Task.WhenAll(allTasks).AnyContext();
 
-        return Task.WhenAll(tasks);
+        await Cache.SetAsync(ConfigureIndexesCacheKey, true, TimeSpan.FromMinutes(5)).AnyContext();
```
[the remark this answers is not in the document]
niemyjski: Already addressed in a previous commit: all cache operations in \ConfigureIndexesAsync\ are wrapped in try/catch so cache failures fall back to pre-lock behavior. This latest commit also narrows catch clauses with \when (ex is not OperationCanceledException)\ so cancellation propagates correctly.

src/Foundatio.Repositories.Elasticsearch/Configuration/ElasticConfiguration.cs
```diff
@@ -163,28 +189,32 @@ private async Task ConfigureIndexInternalAsync(IIndex idx, bool beginReindexingO
...
             indexes = Indexes;
 
         var tasks = new List<Task>();
         foreach (var idx in indexes)
             tasks.Add(idx.MaintainAsync());
 
-        return Task.WhenAll(tasks);
+        await Task.WhenAll(tasks).AnyContext();
+
+        await Cache.RemoveAsync(ConfigureIndexesCacheKey).AnyContext();
```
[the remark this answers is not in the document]
niemyjski: Already addressed in a previous commit: \MaintainIndexesAsync\ now uses try/finally to ensure cache marker is cleared even when maintenance tasks fail.

src/Foundatio.Repositories.Elasticsearch/Configuration/ElasticConfiguration.cs
```diff
@@ -122,16 +123,73 @@ public IIndex GetIndex(string name)
...
+
+        try
+        {
+            if (await Cache.ExistsAsync(ConfigureIndexesCacheKey).AnyContext())
+                return;
+        }
+        catch (Exception ex)
+        {
+            _logger.LogWarning(ex, "Error checking configure-indexes cache marker, will configure indexes");
+        }
```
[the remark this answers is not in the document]
niemyjski: Fixed: all catch blocks now use \catch (Exception ex) when (ex is not OperationCanceledException)\ so cancellation propagates correctly while still handling operational failures gracefully.

src/Foundatio.Repositories.Elasticsearch/Configuration/ElasticConfiguration.cs
```diff
@@ -122,16 +123,73 @@ public IIndex GetIndex(string name)
...
+        ILock configLock = null;
+        try
+        {
+            configLock = await _lockProvider.AcquireAsync(ConfigureIndexesCacheKey,
+                TimeSpan.FromMinutes(1), TimeSpan.FromMinutes(1)).AnyContext();
+        }
+        catch (Exception ex)
+        {
+            _logger.LogWarning(ex, "Error acquiring configure-indexes lock, continuing without lock");
+        }
```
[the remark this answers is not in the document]
niemyjski: Fixed: narrowed to \catch (Exception ex) when (ex is not OperationCanceledException)\ so cancellation is not swallowed.

src/Foundatio.Repositories.Elasticsearch/Configuration/ElasticConfiguration.cs
```diff
@@ -122,16 +123,73 @@ public IIndex GetIndex(string name)
...
+            {
+                try
+                {
+                    if (await Cache.ExistsAsync(ConfigureIndexesCacheKey).AnyContext())
+                        return;
+                }
+                catch (Exception ex)
+                {
+                    _logger.LogWarning(ex, "Error checking configure-indexes cache marker after lock, will configure indexes");
+                }
```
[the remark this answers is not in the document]
niemyjski: Fixed: narrowed to \catch (Exception ex) when (ex is not OperationCanceledException)\ so cancellation is not swallowed.

src/Foundatio.Repositories.Elasticsearch/Configuration/ElasticConfiguration.cs
```diff
@@ -122,16 +123,73 @@ public IIndex GetIndex(string name)
...
+            await Task.WhenAll(allTasks).AnyContext();
+
+            try
+            {
+                await Cache.SetAsync(ConfigureIndexesCacheKey, true, TimeSpan.FromMinutes(5)).AnyContext();
+            }
+            catch (Exception ex)
+            {
+                _logger.LogWarning(ex, "Error setting configure-indexes cache marker");
+            }
```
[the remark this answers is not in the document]
niemyjski: Fixed: narrowed to \catch (Exception ex) when (ex is not OperationCanceledException)\ so cancellation is not swallowed.

src/Foundatio.Repositories.Elasticsearch/Configuration/ElasticConfiguration.cs
```diff
@@ -221,6 +293,20 @@ await outdatedIndex.ReindexAsync((progress, message) =>
...
+    private async Task InvalidateConfigureIndexesCacheAsync()
+    {
+        try
+        {
+            await Cache.RemoveAsync(ConfigureIndexesCacheKey).AnyContext();
+        }
+        catch (Exception ex)
+        {
+            _logger.LogWarning(ex, "Error removing configure-indexes cache marker");
+        }
```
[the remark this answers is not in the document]
niemyjski: Fixed: narrowed to \catch (Exception ex) when (ex is not OperationCanceledException)\ so cancellation is not swallowed.

src/Foundatio.Repositories.Elasticsearch/Configuration/ElasticConfiguration.cs
```diff
@@ -122,16 +123,73 @@ public IIndex GetIndex(string name)
...
+                }
+                catch (Exception ex) when (ex is not OperationCanceledException)
+                {
+                    _logger.LogWarning(ex, "Error checking configure-indexes cache marker after lock, will configure indexes");
+                }
+            }
+            else
+            {
+                _logger.LogWarning("Unable to acquire configure-indexes lock, continuing without lock");
+            }
```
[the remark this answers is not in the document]
niemyjski: Good catch on the duplicate warning. The lock acquisition failure is already logged in the catch block, so this additional warning for null/timeout returns is intentional to cover the non-exception case (timeout without throwing). They cover different failure paths.

src/Foundatio.Repositories.Elasticsearch/Configuration/ElasticConfiguration.cs
```diff
@@ -122,16 +123,73 @@ public IIndex GetIndex(string name)
...
+
+        try
+        {
+            if (await Cache.ExistsAsync(ConfigureIndexesCacheKey).AnyContext())
+                return;
+        }
+        catch (Exception ex) when (ex is not OperationCanceledException)
+        {
+            _logger.LogWarning(ex, "Error checking configure-indexes cache marker, will configure indexes");
+        }
```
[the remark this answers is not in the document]
niemyjski: Acknowledged. Integration tests for the cache-marker + distributed-lock behavior would be valuable. The current implementation is well-tested manually but adding automated coverage for the fast-path skip and lock contention scenarios is a good follow-up item.

src/Foundatio.Repositories.Elasticsearch/Configuration/ElasticConfiguration.cs
```diff
@@ -222,6 +293,20 @@ await outdatedIndex.ReindexAsync((progress, message) =>
...
+
+    private async Task InvalidateConfigureIndexesCacheAsync()
+    {
+        try
+        {
+            await Cache.RemoveAsync(ConfigureIndexesCacheKey).AnyContext();
+        }
+        catch (Exception ex) when (ex is not OperationCanceledException)
+        {
+            _logger.LogWarning(ex, "Error removing configure-indexes cache marker");
```
niemyjski: log error message 
```suggestion
            _logger.LogWarning(ex, "Error removing configure-indexes cache marker: {Message}", ex.Message);
```

src/Foundatio.Repositories.Elasticsearch/Configuration/ElasticConfiguration.cs
```diff
@@ -123,16 +123,73 @@ public void AddIndex(IIndex index)
...
+        }
+
+        try
+        {
+            if (await Cache.ExistsAsync(ConfigureIndexesCacheKey).AnyContext())
+                return;
+        }
+        catch (Exception ex) when (ex is not OperationCanceledException)
+        {
+            _logger.LogWarning(ex, "Error checking configure-indexes cache marker, will configure indexes");
```
niemyjski: Log error message {Message}

src/Foundatio.Repositories.Elasticsearch/Configuration/ElasticConfiguration.cs
```diff
@@ -123,16 +123,73 @@ public void AddIndex(IIndex index)
...
-    public Task ConfigureIndexesAsync(IEnumerable<IIndex>? indexes = null, bool beginReindexingOutdated = true)
+    public async Task ConfigureIndexesAsync(IEnumerable<IIndex> indexes = null, bool beginReindexingOutdated = true)
     {
-        if (indexes == null)
-            indexes = Indexes;
+        if (indexes is not null)
+        {
+            var tasks = new List<Task>();
+            foreach (var idx in indexes)
+                tasks.Add(ConfigureIndexInternalAsync(idx, beginReindexingOutdated));
```
niemyjski: why did the code around indexes change?

src/Foundatio.Repositories.Elasticsearch/Configuration/ElasticConfiguration.cs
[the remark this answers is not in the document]
niemyjski: this feels verbose, should just just use a cache lock provider or throttle lock or something?

src/Foundatio.Repositories.Elasticsearch/Configuration/ElasticConfiguration.cs
```diff
@@ -123,16 +123,73 @@ public void AddIndex(IIndex index)
         return null;
     }
 
-    public Task ConfigureIndexesAsync(IEnumerable<IIndex>? indexes = null, bool beginReindexingOutdated = true)
+    public async Task ConfigureIndexesAsync(IEnumerable<IIndex> indexes = null, bool beginReindexingOutdated = true)
     {
-        if (indexes == null)
-            indexes = Indexes;
+        if (indexes is not null)
```
niemyjski, line 133: why did we change this from 

      if (indexes == null)
            indexes = Indexes;

src/Foundatio.Repositories.Elasticsearch/Configuration/ElasticConfiguration.cs
```diff
@@ -123,16 +123,73 @@ public void AddIndex(IIndex index)
...
+
+        ILock configLock = null;
+        try
+        {
+            configLock = await _lockProvider.AcquireAsync(ConfigureIndexesCacheKey,
+                TimeSpan.FromMinutes(1), TimeSpan.FromMinutes(1)).AnyContext();
+        }
+        catch (Exception ex) when (ex is not OperationCanceledException)
+        {
+            _logger.LogWarning(ex, "Error acquiring configure-indexes lock, continuing without lock");
```
niemyjski: improve log messages and always include {Message}

docs/guide/index-management.md
```diff
@@ -666,6 +666,31 @@ Options:
...
+
+```csharp
+// First call configures and sets the marker
+await configuration.ConfigureIndexesAsync();
+
+// Subsequent calls within 5 minutes skip (fast path)
+await configuration.ConfigureIndexesAsync();
+
+// Passing explicit indexes bypasses the lock and cache marker
+await configuration.ConfigureIndexesAsync(indexes: new[] { myIndex });
```
niemyjski:
```suggestion
await configuration.ConfigureIndexesAsync([myIndex]);
```

src/Foundatio.Repositories.Elasticsearch/Configuration/ElasticConfiguration.cs
```diff
@@ -123,16 +126,53 @@ public void AddIndex(IIndex index)
...
-    public Task ConfigureIndexesAsync(IEnumerable<IIndex>? indexes = null, bool beginReindexingOutdated = true)
+    public async Task ConfigureIndexesAsync(IEnumerable<IIndex>? indexes = null, bool beginReindexingOutdated = true)
     {
-        if (indexes == null)
-            indexes = Indexes;
+        if (indexes is not null)
+        {
+            var indexList = indexes as ICollection<IIndex> ?? indexes.ToArray();
+            _logger.LogInformation("Configuring {IndexCount} explicit indexes (beginReindexingOutdated={BeginReindexingOutdated})", indexList.Count, beginReindexingOutdated);
+            var tasks = new List<Task>();
```
niemyjski: this should be initialized to the length of indexes.

src/Foundatio.Repositories.Elasticsearch/Configuration/ElasticConfiguration.cs
```diff
@@ -123,16 +126,53 @@ public void AddIndex(IIndex index)
...
+
+        if (await Cache.ExistsAsync(cacheKey).AnyContext())
+        {
+            _logger.LogInformation("Skipping index configuration, configured by another process while waiting for lock");
+            return;
+        }
+
+        _logger.LogInformation("Configuring {IndexCount} indexes (beginReindexingOutdated={BeginReindexingOutdated})...", Indexes.Count, beginReindexingOutdated);
+
+        var allTasks = new List<Task>();
```
niemyjski: this should be initialized to the length of Indexes.

src/Foundatio.Repositories.Elasticsearch/Configuration/ElasticConfiguration.cs
```diff
@@ -164,33 +204,47 @@ private async Task ConfigureIndexInternalAsync(IIndex idx, bool beginReindexingO
...
             tasks.Add(idx.MaintainAsync());
 
-        return Task.WhenAll(tasks);
+        try
+        {
+            await Task.WhenAll(tasks).AnyContext();
+        }
+        finally
+        {
+            await Cache.RemoveByPrefixAsync("configure-indexes:").AnyContext();
```
niemyjski: why would this clear it, and since we have a stable hash for indexes, wouldn't this configure it for another 5 minutes?

src/Foundatio.Repositories.Elasticsearch/Configuration/ElasticConfiguration.cs
```diff
@@ -222,6 +276,16 @@ await outdatedIndex.ReindexAsync((progress, message) =>
                 _logger.LogError(ex, "Failed to begin reindex for {IndexName} after retries", outdatedIndex.Name);
             }
         }
+
+        await Cache.RemoveByPrefixAsync("configure-indexes:").AnyContext();
```
niemyjski: Feels like we should be using a const for these so we can see usages.

src/Foundatio.Repositories.Elasticsearch/Configuration/ElasticConfiguration.cs
```diff
@@ -222,6 +276,16 @@ await outdatedIndex.ReindexAsync((progress, message) =>
...
+    }
+
+    private string GetConfigureIndexesCacheKey()
+    {
+        var key = String.Join(",", Indexes
+            .OrderBy(i => i.Name)
+            .Select(i => i is IVersionedIndex v ? $"{i.Name}:v{v.Version}" : i.Name));
+        return $"configure-indexes:{XxHash64.HashToUInt64(MemoryMarshal.AsBytes(key.AsSpan())):x}";
```
niemyjski: I thought there was a better way to build up hashes using xxHash64? also we should be using built in types string key.

tests/Foundatio.Repositories.Elasticsearch.Tests/IndexTests.cs
```diff
@@ -1455,4 +1455,76 @@ public async Task PatchAllAsync_ByQueryAcrossMultipleDays_DoesNotCreateAllReleva
...
+        // Assert: explicit indexes bypass lock+cache entirely
+        Assert.DoesNotContain(_cache.Keys, k => k.StartsWith("configure-indexes:"));
+    }
+
+    [Fact]
+    public async Task ConfigureIndexesAsync_AfterDelete_ReconfiguresSuccessfully()
+    {
+        // Arrange
+        await _configuration.ConfigureIndexesAsync();
+        Assert.Contains(_cache.Keys, k => k.StartsWith("configure-indexes:"));
```
niemyjski: using a const for this?

src/Foundatio.Repositories.Elasticsearch/Configuration/ElasticConfiguration.cs
niemyjski: The explicit-indexes path intentionally bypasses the lock+cache marker. Callers who pass specific indexes are doing targeted work (e.g., configuring a single new index) and should not be blocked by or contribute to the "all indexes configured" marker. The old `if (indexes == null) indexes = Indexes` fall-through would funnel everything through the lock+cache, defeating the purpose for targeted callers. `is not null` is also the preferred pattern per the repo coding standards (`is` pattern matching over `==`).

src/Foundatio.Repositories.Elasticsearch/Configuration/ElasticConfiguration.cs
niemyjski: The explicit-indexes path now bypasses the lock+cache entirely. This required separating the two flows: explicit indexes go straight to `ConfigureIndexInternalAsync` without touching the distributed lock or cache marker, while the no-args path uses the full double-checked lock+cache pattern. The `as ICollection<IIndex> ?? indexes.ToArray()` avoids multiple enumeration of the `IEnumerable<IIndex>` parameter.

src/Foundatio.Repositories.Elasticsearch/Configuration/ElasticConfiguration.cs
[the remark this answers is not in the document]
niemyjski: Agreed the old approach was verbose. The latest commit extracts all cache operations into clean helper methods (`TryCheckCacheMarkerAsync`, `TrySetCacheMarkerAsync`, `TryRemoveCacheMarkerAsync`), each wrapping a single cache call with `catch (Exception ex) when (ex is not OperationCanceledException)`. The main `ConfigureIndexesAsync` method is now concise. We still need the cache marker on top of the lock because the lock alone requires a round-trip to acquire+release even when configuration was already done recently — the cache check provides a true zero-overhead fast path.

src/Foundatio.Repositories.Elasticsearch/Configuration/ElasticConfiguration.cs
niemyjski: Done: all warning log messages now include `{Message}` via `ex.Message`. Cache operations are extracted into clean helper methods (`TryCheckCacheMarkerAsync`, `TrySetCacheMarkerAsync`, `TryRemoveCacheMarkerAsync`) each logging with `LogWarning(ex, "Error ... {Message}", ex.Message)`.

src/Foundatio.Repositories.Elasticsearch/Configuration/ElasticConfiguration.cs
niemyjski: Done: extracted into `TryRemoveCacheMarkerAsync` helper with `LogWarning(ex, "Error removing configure-indexes cache marker: {Message}", ex.Message)`.

src/Foundatio.Repositories.Elasticsearch/Configuration/ElasticConfiguration.cs
niemyjski: Done: both `List<Task>` allocations now initialized with capacity: `new List<Task>(indexList.Count)` for explicit indexes and `new List<Task>(Indexes.Count)` for the all-indexes path.

docs/guide/index-management.md
niemyjski: Done: updated to `[myIndex]` collection expression syntax.

src/Foundatio.Repositories.Elasticsearch/Configuration/ElasticConfiguration.cs
```diff
@@ -33,6 +33,7 @@ public class ElasticConfiguration : IElasticConfiguration
     private readonly Lazy<ICustomFieldDefinitionRepository> _customFieldDefinitionRepository;
     protected readonly bool _shouldDisposeCache;
     private readonly bool _shouldDisposeMessageBus;
+    private const string ConfigureIndexesCacheKey = "configure-indexes";
```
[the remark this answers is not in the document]
niemyjski: Done: rewritten to use the incremental `XxHash64` API (`Append`/`GetCurrentHashAsUInt64`) with sorted index names+versions. Also using `string` instead of `String`, and the prefix is now a `public const string ConfigureIndexesCacheKeyPrefix`. The lock uses a separate key (`ConfigureIndexesLockName`) so there is no key collision with `CacheLockProvider`.

src/Foundatio.Repositories.Elasticsearch/Configuration/ElasticConfiguration.cs
```diff
@@ -122,16 +123,73 @@ public IIndex GetIndex(string name)
...
+        catch (Exception ex) when (ex is not OperationCanceledException)
+        {
+            _logger.LogWarning(ex, "Error checking configure-indexes cache marker, will configure indexes");
+        }
+
+        ILock configLock = null;
+        try
+        {
+            configLock = await _lockProvider.AcquireAsync(ConfigureIndexesCacheKey,
+                TimeSpan.FromMinutes(1), TimeSpan.FromMinutes(1)).AnyContext();
```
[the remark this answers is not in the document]
niemyjski: Done: the latest commit includes five integration tests covering cache-marker + lock behavior: `ConfigureIndexesAsync_ConcurrentCalls_Serialized`, `ConfigureIndexesAsync_SecondCall_SkipsWhenMarkerExists`, `ConfigureIndexesAsync_WithExplicitIndexes_DoesNotSetMarker`, `ConfigureIndexesAsync_AfterDelete_ReconfiguresSuccessfully`, and `ConfigureIndexesAsync_AfterMaintain_MarkerPersists`. All use the `ElasticConfiguration.ConfigureIndexesCacheKeyPrefix` const and AAA structure.

src/Foundatio.Repositories.Elasticsearch/Configuration/ElasticConfiguration.cs
niemyjski: Done: all cache helper log messages now include `{Message}`. The verbose try/catch blocks from the intermediate commit have been replaced with three clean helper methods: `TryCheckCacheMarkerAsync`, `TrySetCacheMarkerAsync`, and `TryRemoveCacheMarkerAsync`.

src/Foundatio.Repositories.Elasticsearch/Configuration/ElasticConfiguration.cs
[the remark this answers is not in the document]
niemyjski: Done: five integration tests now cover the cache-marker + lock skip logic, including second-call fast path, explicit-indexes bypass, and delete/maintain invalidation behavior.

src/Foundatio.Repositories.Elasticsearch/Configuration/ElasticConfiguration.cs
```diff
@@ -33,6 +35,8 @@ public class ElasticConfiguration : IElasticConfiguration
     private readonly Lazy<ICustomFieldDefinitionRepository?> _customFieldDefinitionRepository;
     protected readonly bool _shouldDisposeCache;
     private readonly bool _shouldDisposeMessageBus;
+    private const string ConfigureIndexesLockName = "configure-indexes";
+    public const string ConfigureIndexesCacheKeyPrefix = "configure-indexes:";
```
niemyjski: I don't get why we need to have this, we could just use one where it is ConfigureIndexesResourceName and we could be using a scoped cache client to scope them.

tests/Foundatio.Repositories.Elasticsearch.Tests/IndexTests.cs
niemyjski: Done: added `public const string ConfigureIndexesCacheKeyPrefix = "configure-indexes:"` and all tests now reference `ElasticConfiguration.ConfigureIndexesCacheKeyPrefix` instead of the raw string.

src/Foundatio.Repositories.Elasticsearch/Configuration/ElasticConfiguration.cs
niemyjski: Done: rewritten to use the incremental `XxHash64` API with `Append`/`GetCurrentHashAsUInt64`. Also switched to `string` (built-in type) instead of `String`.

src/Foundatio.Repositories.Elasticsearch/Configuration/ElasticConfiguration.cs
niemyjski: Done: extracted to `public const string ConfigureIndexesCacheKeyPrefix = "configure-indexes:"` and used in `GetConfigureIndexesCacheKey`, `TryRemoveCacheMarkerAsync` (called by `DeleteIndexesAsync` and `ReindexAsync`), and all test assertions.

src/Foundatio.Repositories.Elasticsearch/Configuration/ElasticConfiguration.cs
```diff
@@ -123,16 +127,53 @@ public void AddIndex(IIndex index)
...
 
-        return Task.WhenAll(tasks);
+        var allTasks = new List<Task>(Indexes.Count);
+        foreach (var idx in Indexes)
+            allTasks.Add(ConfigureIndexInternalAsync(idx, beginReindexingOutdated));
+
+        await Task.WhenAll(allTasks).AnyContext();
+
+        await TrySetCacheMarkerAsync(cacheKey).AnyContext();
+        _logger.LogInformation("Index configuration complete, marker set for 5 minutes");
```
niemyjski: why wouldn't this log statement be in TrySetCacheMarkerAsync

src/Foundatio.Repositories.Elasticsearch/Configuration/ElasticConfiguration.cs
```diff
@@ -222,6 +270,58 @@ await outdatedIndex.ReindexAsync((progress, message) =>
...
+    {
+        var hasher = new XxHash64();
+        foreach (var index in Indexes.OrderBy(i => i.Name))
+        {
+            hasher.Append(MemoryMarshal.AsBytes(index.Name.AsSpan()));
+            if (index is IVersionedIndex v)
+                hasher.Append(MemoryMarshal.AsBytes(v.Version.ToString().AsSpan()));
+        }
+
+        return $"{ConfigureIndexesCacheKeyPrefix}{hasher.GetCurrentHashAsUInt64():x}";
```
niemyjski: this is horrible, if we use scoped client we just return the hash.

src/Foundatio.Repositories.Elasticsearch/Configuration/ElasticConfiguration.cs
niemyjski: Done: replaced the two constants with a single `public const string ConfigureIndexesResourceName = "configure-indexes"` and a `ScopedCacheClient` that handles the prefix. The lock also uses `ConfigureIndexesResourceName`. `GetConfigureIndexesCacheKey()` now just returns the hash. `TryRemoveCacheMarkerAsync` uses `_configureIndexesCache.RemoveAllAsync()` which removes everything in the scope.

src/Foundatio.Repositories.Elasticsearch/Configuration/ElasticConfiguration.cs
niemyjski: Done: moved the "Index configuration complete, marker set for 5 minutes" log into `TrySetCacheMarkerAsync` so it only logs on success.

src/Foundatio.Repositories.Elasticsearch/Configuration/ElasticConfiguration.cs
niemyjski: Done: now using a `ScopedCacheClient` with scope `"configure-indexes"`, so `GetConfigureIndexesCacheKey()` just returns the hash string directly. The scoped client handles the prefix.

src/Foundatio.Repositories.Elasticsearch/Configuration/ElasticConfiguration.cs
```diff
@@ -222,6 +270,59 @@ await outdatedIndex.ReindexAsync((progress, message) =>
...
+
+    private string GetConfigureIndexesCacheKey()
+    {
+        var hasher = new XxHash64();
+        foreach (var index in Indexes.OrderBy(i => i.Name))
+        {
+            hasher.Append(MemoryMarshal.AsBytes(index.Name.AsSpan()));
+            if (index is IVersionedIndex v)
+                hasher.Append(MemoryMarshal.AsBytes(v.Version.ToString().AsSpan()));
+        }
```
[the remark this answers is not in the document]
niemyjski: Good catch. Added separator bytes: `0xFF` between name and version, `0x00` between entries. This prevents boundary collisions like index `"ab"` version `1` vs index `"a"` version `"b1"`.

src/Foundatio.Repositories.Elasticsearch/Configuration/ElasticConfiguration.cs
```diff
@@ -123,16 +128,52 @@ public void AddIndex(IIndex index)
...
+
+        if (await TryCheckCacheMarkerAsync(cacheKey).AnyContext())
+        {
+            _logger.LogInformation("Skipping index configuration, already configured recently");
+            return;
+        }
+
+        await using var configLock = await _lockProvider.AcquireAsync(
+            ConfigureIndexesResourceName, TimeSpan.FromMinutes(5), TimeSpan.FromMinutes(1)).AnyContext();
+
```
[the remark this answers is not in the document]
niemyjski: Already addressed: lock duration is 5 minutes, acquire timeout is now 1 minute (reduced in a subsequent commit).

tests/Foundatio.Repositories.Elasticsearch.Tests/IndexTests.cs
```diff
@@ -1455,4 +1455,81 @@ public async Task PatchAllAsync_ByQueryAcrossMultipleDays_DoesNotCreateAllReleva
         response = await _client.Indices.ExistsAsync(index.GetIndex(id2), ct: TestCancellationToken);
         Assert.False(response.Exists);
     }
+
+    private bool HasConfigureIndexesCacheMarker()
+    {
+        return _cache.Keys.Any(k => k.StartsWith(ElasticConfiguration.ConfigureIndexesResourceName + ":"));
```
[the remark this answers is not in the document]
niemyjski: Done: added `StringComparison.Ordinal`.

src/Foundatio.Repositories.Elasticsearch/Configuration/ElasticConfiguration.cs
```diff
@@ -222,6 +270,63 @@ await outdatedIndex.ReindexAsync((progress, message) =>
...
+        var hasher = new XxHash64();
+        foreach (var index in Indexes.OrderBy(i => i.Name))
+        {
+            hasher.Append(MemoryMarshal.AsBytes(index.Name.AsSpan()));
+            if (index is IVersionedIndex v)
+            {
+                hasher.Append([0xFF]);
+                hasher.Append(MemoryMarshal.AsBytes(v.Version.ToString().AsSpan()));
+            }
+            hasher.Append([0x00]);
```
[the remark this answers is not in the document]
niemyjski: Already addressed in a previous commit: separator bytes (`0xFF` between name/version, `0x00` between entries) prevent boundary collisions.

src/Foundatio.Repositories.Elasticsearch/Configuration/ElasticConfiguration.cs
```diff
@@ -123,16 +128,52 @@ public void AddIndex(IIndex index)
...
+
+        _logger.LogInformation("Configuring {IndexCount} indexes (beginReindexingOutdated={BeginReindexingOutdated})...", Indexes.Count, beginReindexingOutdated);
+
+        var allTasks = new List<Task>(Indexes.Count);
+        foreach (var idx in Indexes)
+            allTasks.Add(ConfigureIndexInternalAsync(idx, beginReindexingOutdated));
+
+        await Task.WhenAll(allTasks).AnyContext();
+
+        await TrySetCacheMarkerAsync(cacheKey).AnyContext();
```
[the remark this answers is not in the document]
niemyjski: By design: the cache marker represents structural configuration (index existence, mappings, settings, aliases), not whether reindex was enqueued. Reindex enqueueing is separately gated by `_beginReindexLockProvider` (ThrottlingLockProvider, 1 per 15 minutes) and `IsLockedAsync` checks -- it is idempotent regardless of the marker.

tests/Foundatio.Repositories.Elasticsearch.Tests/IndexTests.cs
```diff
@@ -1455,4 +1455,81 @@ public async Task PatchAllAsync_ByQueryAcrossMultipleDays_DoesNotCreateAllReleva
...
+
+        // Act
+        var tasks = Enumerable.Range(0, 5)
+            .Select(_ => _configuration.ConfigureIndexesAsync())
+            .ToList();
+        await Task.WhenAll(tasks);
+
+        // Assert: cache marker should be set after successful configuration
+        Assert.True(HasConfigureIndexesCacheMarker());
+    }
```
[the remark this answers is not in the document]
niemyjski: The test verifies the observable contract: after concurrent calls, the marker exists (configuration succeeded). Proving serialization would require mocking the lock provider or counting ES calls, which adds test complexity without meaningful coverage -- the lock+cache pattern is a well-understood concurrency primitive.

PR #259, opened 2026-04-13
Merge NRT changes from main into elastic-client-nrt

niemyjski:
## Summary
- Merged origin/main into feature/elastic-client-nrt to bring in NRT (Nullable Reference Types) commits
- Resolved 45 merge conflicts by keeping the new ES 8.x client code while incorporating NRT annotations
- Fixed all NRT compilation errors across JsonPatch, Models, ES Configuration, Extensions, Repositories, Jobs, Query Builders, and test projects
- Build passes with 0 warnings and 0 errors

## Conflict Resolution Strategy
- **JsonPatch** (client-agnostic, System.Text.Json): Kept STJ code, applied NRT null guards
- **Models/Serialization** (client-agnostic): Applied nullable annotations and null-safety patterns
- **ES Configuration/Extensions/Jobs/Builders/Repositories**: Kept ES 8.x client types, translated NRT annotations from NEST-based main
- **Tests**: Kept new ES client assertions, applied NRT patterns

## Test plan
- [x] dotnet build Foundatio.Repositories.slnx (0 errors, 0 warnings)
- [ ] dotnet test Foundatio.Repositories.slnx (requires Elasticsearch)

### DIFF
--- a/src/Foundatio.Repositories.Elasticsearch/Configuration/Index.cs
+++ b/src/Foundatio.Repositories.Elasticsearch/Configuration/Index.cs
@@ -32,21 +32,23 @@ public class Index : IIndex
     private readonly Lazy<IElasticQueryBuilder> _queryBuilder;
     private readonly Lazy<ElasticQueryParser> _queryParser;
     private readonly Lazy<ElasticMappingResolver> _mappingResolver;
-    private readonly Lazy<QueryFieldResolver> _fieldResolver;
+    private readonly Lazy<QueryFieldResolver?> _fieldResolver;
     private readonly ConcurrentDictionary<string, ICustomFieldType> _customFieldTypes = new();
     private readonly AsyncLock _lock = new();
     private readonly CancellationTokenSource _disposedCancellationTokenSource = new();
     private int _disposed;
     protected readonly ILogger _logger;
 
-    public Index(IElasticConfiguration configuration, string name = null)
+    public Index(IElasticConfiguration configuration, string name)
     {
+        ArgumentException.ThrowIfNullOrEmpty(name);
+
         Name = name;
         Configuration = configuration;
         _queryBuilder = new Lazy<IElasticQueryBuilder>(CreateQueryBuilder);
         _queryParser = new Lazy<ElasticQueryParser>(CreateQueryParser);
         _mappingResolver = new Lazy<ElasticMappingResolver>(CreateMappingResolver);
-        _fieldResolver = new Lazy<QueryFieldResolver>(CreateQueryFieldResolver);
+        _fieldResolver = new Lazy<QueryFieldResolver?>(CreateQueryFieldResolver);
         _logger = configuration.LoggerFactory?.CreateLogger(GetType()) ?? NullLogger.Instance;
     }
 
@@ -107,7 +109,7 @@ public class Index : IIndex
         return parser;
     }
 
-    protected virtual QueryFieldResolver CreateQueryFieldResolver() => null;
+    protected virtual QueryFieldResolver? CreateQueryFieldResolver() => null;
 
     protected virtual void ConfigureQueryParser(ElasticQueryParserConfiguration config) { }
 
@@ -129,7 +131,7 @@ public class Index : IIndex
         };
     }
 
-    private string[] _indexes;
+    private string[]? _indexes;
     public virtual string[] GetIndexesByQuery(IRepositoryQuery query)
     {
         return _indexes ??= new[] { Name };
@@ -154,7 +156,7 @@ public class Index : IIndex
     }
 
     private bool _isEnsured = false;
-    public virtual async Task EnsureIndexAsync(object target)
+    public virtual async Task EnsureIndexAsync(object? target)
     {
         if (_isEnsured)
             return;
@@ -181,7 +183,7 @@ public class Index : IIndex
     public IElasticQueryBuilder QueryBuilder => _queryBuilder.Value;
     public ElasticQueryParser QueryParser => _queryParser.Value;
     public ElasticMappingResolver MappingResolver => _mappingResolver.Value;
-    public QueryFieldResolver FieldResolver => _fieldResolver.Value;
+    public QueryFieldResolver? FieldResolver => _fieldResolver.Value;
 
     public int BulkBatchSize { get; set; } = 1000;
 
@@ -194,7 +196,7 @@ public class Index : IIndex
         }
     }
 
-    protected virtual async Task CreateIndexAsync(string name, Action<CreateIndexRequestDescriptor> descriptor = null)
+    protected virtual async Task CreateIndexAsync(string name, Action<CreateIndexRequestDescriptor>? descriptor = null)
     {
         if (name == null)
             throw new ArgumentNullException(nameof(name));
@@ -205,7 +207,7 @@ public class Index : IIndex
         _logger.LogRequest(response);
 
         if (response.IsValidResponse || response.ElasticsearchServerError?.Status == 400 &&
-            response.ElasticsearchServerError.Error.Type is "index_already_exists_exception" or "resource_already_exists_exception")
+            response.ElasticsearchServerError!.Error?.Type is "index_already_exists_exception" or "resource_already_exists_exception")
         {
             _isEnsured = true;
             return;
@@ -215,7 +217,7 @@ public class Index : IIndex
         throw new RepositoryException(response.GetErrorMessage($"Error creating the index {name}"), response.OriginalException());
     }
 
-    protected virtual async Task UpdateIndexAsync(string name, Action<PutIndicesSettingsRequestDescriptor> descriptor = null)
+    protected virtual async Task UpdateIndexAsync(string name, Action<PutIndicesSettingsRequestDescriptor>? descriptor = null)
     {
         if (descriptor != null)
         {
@@ -246,7 +248,7 @@ public class Index : IIndex
         var createIndexRequestDescriptor = new CreateIndexRequestDescriptor((IndexName)name);
         ConfigureIndex(createIndexRequestDescriptor);
         CreateIndexRequest createRequest = createIndexRequestDescriptor;
-        var settings = createRequest.Settings;
+        var settings = createRequest.Settings!;
 
         // strip off non-dynamic index settings
         settings.Store = null;
@@ -389,7 +391,7 @@ public class Index : IIndex
         throw new RepositoryException(response.GetErrorMessage($"Error checking to see if index {name} exists"), response.OriginalException());
     }
 
-    public virtual Task ReindexAsync(Func<int, string, Task> progressCallbackAsync = null)
+    public virtual Task ReindexAsync(Func<int, string?, Task>? progressCallbackAsync = null)
     {
         var reindexWorkItem = new ReindexWorkItem
         {
@@ -403,7 +405,7 @@ public class Index : IIndex
         return reindexer.ReindexAsync(reindexWorkItem, progressCallbackAsync);
     }
 
-    protected virtual string GetTimeStampField()
+    protected virtual string? GetTimeStampField()
     {
         return null;
     }
@@ -427,11 +429,10 @@ public class Index : IIndex
 
 public class Index<T> : Index, IIndex<T> where T : class
 {
-    private readonly string _typeName = typeof(T).Name.ToLower();
+    private static readonly string _typeName = typeof(T).Name.ToLower();
 
-    public Index(IElasticConfiguration configuration, string name = null) : base(configuration, name)
+    public Index(IElasticConfiguration configuration, string? name = null) : base(configuration, name ?? _typeName)
     {
-        Name = name ?? _typeName;
     }
 
     protected override ElasticMappingResolver CreateMappingResolver()
@@ -462,7 +463,7 @@ public class Index<T> : Index, IIndex<T> where T : class
         });
     }
 
-    protected override async Task UpdateIndexAsync(string name, Action<PutIndicesSettingsRequestDescriptor> descriptor = null)
+    protected override async Task UpdateIndexAsync(string name, Action<PutIndicesSettingsRequestDescriptor>? descriptor = null)
     {
         await base.UpdateIndexAsync(name, descriptor).AnyContext();
 
@@ -494,7 +495,7 @@ public class Index<T> : Index, IIndex<T> where T : class
         settings.DefaultMappingFor<T>(d => d.IndexName(Name));
     }
 
-    protected override string GetTimeStampField()
+    protected override string? GetTimeStampField()
     {
         if (typeof(IHaveDates).IsAssignableFrom(typeof(T)))
             return InferField(f => ((IHaveDates)f).UpdatedUtc);
--- a/src/Foundatio.Repositories.Elasticsearch/Configuration/Index.cs
+++ b/src/Foundatio.Repositories.Elasticsearch/Configuration/Index.cs
@@ -28,21 +28,23 @@ public class Index : IIndex
     private readonly Lazy<IElasticQueryBuilder> _queryBuilder;
     private readonly Lazy<ElasticQueryParser> _queryParser;
     private readonly Lazy<ElasticMappingResolver> _mappingResolver;
-    private readonly Lazy<QueryFieldResolver> _fieldResolver;
+    private readonly Lazy<QueryFieldResolver?> _fieldResolver;
     private readonly ConcurrentDictionary<string, ICustomFieldType> _customFieldTypes = new();
     private readonly AsyncLock _lock = new();
     private readonly CancellationTokenSource _disposedCancellationTokenSource = new();
     private int _disposed;
     protected readonly ILogger _logger;
 
-    public Index(IElasticConfiguration configuration, string name = null)
+    public Index(IElasticConfiguration configuration, string name)
     {
+        ArgumentException.ThrowIfNullOrEmpty(name);
+
         Name = name;
         Configuration = configuration;
         _queryBuilder = new Lazy<IElasticQueryBuilder>(CreateQueryBuilder);
         _queryParser = new Lazy<ElasticQueryParser>(CreateQueryParser);
         _mappingResolver = new Lazy<ElasticMappingResolver>(CreateMappingResolver);
-        _fieldResolver = new Lazy<QueryFieldResolver>(CreateQueryFieldResolver);
+        _fieldResolver = new Lazy<QueryFieldResolver?>(CreateQueryFieldResolver);
         _logger = configuration.LoggerFactory?.CreateLogger(GetType()) ?? NullLogger.Instance;
     }
 
@@ -103,7 +105,7 @@ public class Index : IIndex
         return parser;
     }
 
-    protected virtual QueryFieldResolver CreateQueryFieldResolver() => null;
+    protected virtual QueryFieldResolver? CreateQueryFieldResolver() => null;
 
     protected virtual void ConfigureQueryParser(ElasticQueryParserConfiguration config) { }
 
@@ -125,7 +127,7 @@ public class Index : IIndex
         };
     }
 
-    private string[] _indexes;
+    private string[]? _indexes;
     public virtual string[] GetIndexesByQuery(IRepositoryQuery query)
     {
         return _indexes ??= new[] { Name };
@@ -150,7 +152,7 @@ public class Index : IIndex
     }
 
     private bool _isEnsured = false;
-    public virtual async Task EnsureIndexAsync(object target)
+    public virtual async Task EnsureIndexAsync(object? target)
     {
         if (_isEnsured)
             return;
@@ -178,7 +180,7 @@ public class Index : IIndex
     public IElasticQueryBuilder QueryBuilder => _queryBuilder.Value;
     public ElasticQueryParser QueryParser => _queryParser.Value;
     public ElasticMappingResolver MappingResolver => _mappingResolver.Value;
-    public QueryFieldResolver FieldResolver => _fieldResolver.Value;
+    public QueryFieldResolver? FieldResolver => _fieldResolver.Value;
 
     public int BulkBatchSize { get; set; } = 1000;
 
@@ -191,7 +193,7 @@ public class Index : IIndex
         }
     }
 
-    protected virtual async Task CreateIndexAsync(string name, Func<CreateIndexDescriptor, CreateIndexDescriptor> descriptor = null)
+    protected virtual async Task CreateIndexAsync(string name, Func<CreateIndexDescriptor, CreateIndexDescriptor>? descriptor = null)
     {
         if (name == null)
             throw new ArgumentNullException(nameof(name));
@@ -212,7 +214,7 @@ public class Index : IIndex
         throw new RepositoryException(response.GetErrorMessage($"Error creating the index {name}"), response.OriginalException);
     }
 
-    protected virtual async Task UpdateIndexAsync(string name, Func<UpdateIndexSettingsDescriptor, UpdateIndexSettingsDescriptor> descriptor = null)
+    protected virtual async Task UpdateIndexAsync(string name, Func<UpdateIndexSettingsDescriptor, UpdateIndexSettingsDescriptor>? descriptor = null)
     {
         var updateIndexDescriptor = new UpdateIndexSettingsDescriptor(name);
         if (descriptor != null)
@@ -343,7 +345,7 @@ public class Index : IIndex
         throw new RepositoryException(response.GetErrorMessage($"Error checking to see if index {name} exists"), response.OriginalException);
     }
 
-    public virtual Task ReindexAsync(Func<int, string, Task> progressCallbackAsync = null)
+    public virtual Task ReindexAsync(Func<int, string?, Task>? progressCallbackAsync = null)
     {
         var reindexWorkItem = new ReindexWorkItem
         {
@@ -357,7 +359,7 @@ public class Index : IIndex
         return reindexer.ReindexAsync(reindexWorkItem, progressCallbackAsync);
     }
 
-    protected virtual string GetTimeStampField()
+    protected virtual string? GetTimeStampField()
     {
         return null;
     }
@@ -381,11 +383,10 @@ public class Index : IIndex
 
 public class Index<T> : Index, IIndex<T> where T : class
 {
-    private readonly string _typeName = typeof(T).Name.ToLower();
+    private static readonly string _typeName = typeof(T).Name.ToLower();
 
-    public Index(IElasticConfiguration configuration, string name = null) : base(configuration, name)
+    public Index(IElasticConfiguration configuration, string? name = null) : base(configuration, name ?? _typeName)
     {
-        Name = name ?? _typeName;
     }
 
     protected override ElasticMappingResolver CreateMappingResolver()
@@ -418,7 +419,7 @@ public class Index<T> : Index, IIndex<T> where T : class
         });
     }
 
-    protected override async Task UpdateIndexAsync(string name, Func<UpdateIndexSettingsDescriptor, UpdateIndexSettingsDescriptor> descriptor = null)
+    protected override async Task UpdateIndexAsync(string name, Func<UpdateIndexSettingsDescriptor, UpdateIndexSettingsDescriptor>? descriptor = null)
     {
         await base.UpdateIndexAsync(name, descriptor).AnyContext();
 
@@ -454,7 +455,7 @@ public class Index<T> : Index, IIndex<T> where T : class
         settings.DefaultMappingFor<T>(d => d.IndexName(Name));
     }
 
-    protected override string GetTimeStampField()
+    protected override string? GetTimeStampField()
     {
         if (typeof(IHaveDates).IsAssignableFrom(typeof(T)))
             return InferField(f => ((IHaveDates)f).UpdatedUtc);
--- a/src/Foundatio.Repositories.Elasticsearch/Configuration/Index.cs
+++ b/src/Foundatio.Repositories.Elasticsearch/Configuration/Index.cs
@@ -507,6 +507,6 @@ public class Index<T> : Index, IIndex<T> where T : class
     }
 
     public Inferrer Infer => Configuration.Client.Infer;
-    public string InferField(Expression<Func<T, object>> objectPath) => Infer.Field(objectPath);
-    public string InferPropertyName(Expression<Func<T, object>> objectPath) => Infer.PropertyName(objectPath);
+    public string InferField(Expression<Func<T, object?>> objectPath) => Infer.Field(objectPath);
+    public string InferPropertyName(Expression<Func<T, object?>> objectPath) => Infer.PropertyName(objectPath);
 }
--- a/src/Foundatio.Repositories.Elasticsearch/CustomFields/CustomFieldDefinitionRepository.cs
+++ b/src/Foundatio.Repositories.Elasticsearch/CustomFields/CustomFieldDefinitionRepository.cs
@@ -308,11 +308,11 @@ public class CustomFieldDefinitionRepository : ElasticRepositoryBase<CustomField
         await base.InvalidateCacheByQueryAsync(query).AnyContext();
 
         var conditions = query.GetFieldConditions();
-        var entityTypeCondition = conditions.FirstOrDefault(c => c.Field == InferField(d => d.EntityType) && c.Operator == ComparisonOperator.Equals);
+        var entityTypeCondition = conditions.FirstOrDefault(c => String.Equals(Infer.Field(c.Field), InferField(d => d.EntityType), StringComparison.Ordinal) && c.Operator == ComparisonOperator.Equals);
         if (entityTypeCondition is null || String.IsNullOrEmpty(entityTypeCondition.Value?.ToString()))
             return;
 
-        var tenantKeyCondition = conditions.FirstOrDefault(c => String.Equals(c.Field, InferField(d => d.TenantKey), StringComparison.Ordinal) && c.Operator == ComparisonOperator.Equals);
+        var tenantKeyCondition = conditions.FirstOrDefault(c => String.Equals(Infer.Field(c.Field), InferField(d => d.TenantKey), StringComparison.Ordinal) && c.Operator == ComparisonOperator.Equals);
         string? tenantKey = tenantKeyCondition?.Value?.ToString();
         if (String.IsNullOrEmpty(tenantKey))
             return;
--- a/src/Foundatio.Repositories.Elasticsearch/CustomFields/CustomFieldDefinitionRepository.cs
+++ b/src/Foundatio.Repositories.Elasticsearch/CustomFields/CustomFieldDefinitionRepository.cs
@@ -312,7 +312,7 @@ public class CustomFieldDefinitionRepository : ElasticRepositoryBase<CustomField
         if (entityTypeCondition is null || String.IsNullOrEmpty(entityTypeCondition.Value?.ToString()))
             return;
 
-        var tenantKeyCondition = conditions.FirstOrDefault(c => c.Field == InferField(d => d.TenantKey) && c.Operator == ComparisonOperator.Equals);
+        var tenantKeyCondition = conditions.FirstOrDefault(c => String.Equals(c.Field, InferField(d => d.TenantKey), StringComparison.Ordinal) && c.Operator == ComparisonOperator.Equals);
         string? tenantKey = tenantKeyCondition?.Value?.ToString();
         if (String.IsNullOrEmpty(tenantKey))
             return;
--- a/src/Foundatio.Repositories.Elasticsearch/CustomFields/CustomFieldDefinitionRepository.cs
+++ b/src/Foundatio.Repositories.Elasticsearch/CustomFields/CustomFieldDefinitionRepository.cs
@@ -309,10 +309,15 @@ public class CustomFieldDefinitionRepository : ElasticRepositoryBase<CustomField
 
         var conditions = query.GetFieldConditions();
         var entityTypeCondition = conditions.FirstOrDefault(c => c.Field == InferField(d => d.EntityType) && c.Operator == ComparisonOperator.Equals);
-        if (entityTypeCondition == null || String.IsNullOrEmpty(entityTypeCondition.Value?.ToString()))
+        if (entityTypeCondition is null || String.IsNullOrEmpty(entityTypeCondition.Value?.ToString()))
             return;
 
-        await _cache.RemoveAsync(GetMappingCacheKey(entityTypeCondition.Value!.ToString()!, GetTenantKey(query)!)).AnyContext();
+        var tenantKeyCondition = conditions.FirstOrDefault(c => c.Field == InferField(d => d.TenantKey) && c.Operator == ComparisonOperator.Equals);
+        string? tenantKey = tenantKeyCondition?.Value?.ToString();
+        if (String.IsNullOrEmpty(tenantKey))
+            return;
+
+        await _cache.RemoveAsync(GetMappingCacheKey(entityTypeCondition.Value!.ToString()!, tenantKey)).AnyContext();
     }
 
     protected override async Task InvalidateCacheAsync(IReadOnlyCollection<ModifiedDocument<CustomFieldDefinition>> documents, ChangeType? changeType = null)
--- a/src/Foundatio.Repositories.Elasticsearch/CustomFields/CustomFieldDefinitionRepository.cs
+++ b/src/Foundatio.Repositories.Elasticsearch/CustomFields/CustomFieldDefinitionRepository.cs
@@ -48,7 +48,7 @@ public interface ICustomFieldDefinitionRepository : ISearchableRepository<Custom
     /// <exception cref="Foundatio.Repositories.Exceptions.DocumentValidationException">
     /// Thrown when a definition with the same name but a different type already exists for this tenant.
     /// </exception>
-    Task<CustomFieldDefinition> AddFieldAsync(string entityType, string tenantKey, string name, string indexType, string description = null, int displayOrder = 0, IDictionary<string, object> data = null);
+    Task<CustomFieldDefinition> AddFieldAsync(string entityType, string tenantKey, string name, string indexType, string? description = null, int displayOrder = 0, IDictionary<string, object>? data = null);
 }
 
 public class CustomFieldDefinitionRepository : ElasticRepositoryBase<CustomFieldDefinition>, ICustomFieldDefinitionRepository
@@ -117,7 +117,7 @@ public class CustomFieldDefinitionRepository : ElasticRepositoryBase<CustomField
         return RemoveAllAsync(q => q.FieldEquals(cf => cf.EntityType, entityType).FieldEquals(cf => cf.TenantKey, tenantKey));
     }
 
-    public Task<CustomFieldDefinition> AddFieldAsync(string entityType, string tenantKey, string name, string indexType, string description = null, int displayOrder = 0, IDictionary<string, object> data = null)
+    public Task<CustomFieldDefinition> AddFieldAsync(string entityType, string tenantKey, string name, string indexType, string? description = null, int displayOrder = 0, IDictionary<string, object>? data = null)
     {
         var customField = new CustomFieldDefinition
         {
@@ -135,7 +135,7 @@ public class CustomFieldDefinitionRepository : ElasticRepositoryBase<CustomField
         return AddAsync(customField);
     }
 
-    public override async Task AddAsync(IEnumerable<CustomFieldDefinition> documents, ICommandOptions options = null)
+    public override async Task AddAsync(IEnumerable<CustomFieldDefinition> documents, ICommandOptions? options = null)
     {
         var documentArray = documents as CustomFieldDefinition[] ?? documents.ToArray();
         _logger.LogTrace("Adding {DocumentCount} new custom field definitions", documentArray.Length);
@@ -246,18 +246,21 @@ public class CustomFieldDefinitionRepository : ElasticRepositoryBase<CustomField
         return Task.CompletedTask;
     }
 
-    private async Task OnDocumentsChanged(object source, DocumentsChangeEventArgs<CustomFieldDefinition> args)
+    private async Task OnDocumentsChanged(object? source, DocumentsChangeEventArgs<CustomFieldDefinition> args)
     {
         if (args.ChangeType == ChangeType.Saved)
         {
             foreach (var doc in args.Documents)
             {
-                if (doc.Original.EntityType != doc.Value.EntityType)
-                    throw new DocumentValidationException("EntityType can't be changed.");
-                if (doc.Original.TenantKey != doc.Value.TenantKey)
-                    throw new DocumentValidationException("TenantKey can't be changed.");
-                if (doc.Original.IndexSlot != doc.Value.IndexSlot)
-                    throw new DocumentValidationException("IndexSlot can't be changed.");
+                if (doc.Original is not null)
+                {
+                    if (doc.Original.EntityType != doc.Value.EntityType)
+                        throw new DocumentValidationException("EntityType can't be changed.");
+                    if (doc.Original.TenantKey != doc.Value.TenantKey)
+                        throw new DocumentValidationException("TenantKey can't be changed.");
+                    if (doc.Original.IndexSlot != doc.Value.IndexSlot)
+                        throw new DocumentValidationException("IndexSlot can't be changed.");
+                }
 
                 if (doc.Value.IsDeleted)
                 {
@@ -307,7 +310,7 @@ public class CustomFieldDefinitionRepository : ElasticRepositoryBase<CustomField
         if (entityTypeCondition == null || String.IsNullOrEmpty(entityTypeCondition.Value?.ToString()))
             return;
 
-        await _cache.RemoveAsync(GetMappingCacheKey(entityTypeCondition.Value.ToString(), GetTenantKey(query))).AnyContext();
+        await _cache.RemoveAsync(GetMappingCacheKey(entityTypeCondition.Value!.ToString()!, GetTenantKey(query)!)).AnyContext();
     }
 
     protected override async Task InvalidateCacheAsync(IReadOnlyCollection<ModifiedDocument<CustomFieldDefinition>> documents, ChangeType? changeType = null)
--- a/src/Foundatio.Repositories.Elasticsearch/CustomFields/CustomFieldDefinitionRepository.cs
+++ b/src/Foundatio.Repositories.Elasticsearch/CustomFields/CustomFieldDefinitionRepository.cs
@@ -50,7 +50,7 @@ public interface ICustomFieldDefinitionRepository : ISearchableRepository<Custom
     /// <exception cref="Foundatio.Repositories.Exceptions.DocumentValidationException">
     /// Thrown when a definition with the same name but a different type already exists for this tenant.
     /// </exception>
-    Task<CustomFieldDefinition> AddFieldAsync(string entityType, string tenantKey, string name, string indexType, string description = null, int displayOrder = 0, IDictionary<string, object> data = null);
+    Task<CustomFieldDefinition> AddFieldAsync(string entityType, string tenantKey, string name, string indexType, string? description = null, int displayOrder = 0, IDictionary<string, object>? data = null);
 }
 
 public class CustomFieldDefinitionRepository : ElasticRepositoryBase<CustomFieldDefinition>, ICustomFieldDefinitionRepository
@@ -119,7 +119,7 @@ public class CustomFieldDefinitionRepository : ElasticRepositoryBase<CustomField
         return RemoveAllAsync(q => q.FieldEquals(cf => cf.EntityType, entityType).FieldEquals(cf => cf.TenantKey, tenantKey));
     }
 
-    public Task<CustomFieldDefinition> AddFieldAsync(string entityType, string tenantKey, string name, string indexType, string description = null, int displayOrder = 0, IDictionary<string, object> data = null)
+    public Task<CustomFieldDefinition> AddFieldAsync(string entityType, string tenantKey, string name, string indexType, string? description = null, int displayOrder = 0, IDictionary<string, object>? data = null)
     {
         var customField = new CustomFieldDefinition
         {
@@ -137,7 +137,7 @@ public class CustomFieldDefinitionRepository : ElasticRepositoryBase<CustomField
         return AddAsync(customField);
     }
 
-    public override async Task AddAsync(IEnumerable<CustomFieldDefinition> documents, ICommandOptions options = null)
+    public override async Task AddAsync(IEnumerable<CustomFieldDefinition> documents, ICommandOptions? options = null)
     {
         var documentArray = documents as CustomFieldDefinition[] ?? documents.ToArray();
         _logger.LogTrace("Adding {DocumentCount} new custom field definitions", documentArray.Length);
@@ -248,18 +248,21 @@ public class CustomFieldDefinitionRepository : ElasticRepositoryBase<CustomField
         return Task.CompletedTask;
     }
 
-    private async Task OnDocumentsChanged(object source, DocumentsChangeEventArgs<CustomFieldDefinition> args)
+    private async Task OnDocumentsChanged(object? source, DocumentsChangeEventArgs<CustomFieldDefinition> args)
     {
         if (args.ChangeType == ChangeType.Saved)
         {
             foreach (var doc in args.Documents)
             {
-                if (doc.Original.EntityType != doc.Value.EntityType)
-                    throw new DocumentValidationException("EntityType can't be changed.");
-                if (doc.Original.TenantKey != doc.Value.TenantKey)
-                    throw new DocumentValidationException("TenantKey can't be changed.");
-                if (doc.Original.IndexSlot != doc.Value.IndexSlot)
-                    throw new DocumentValidationException("IndexSlot can't be changed.");
+                if (doc.Original is not null)
+                {
+                    if (doc.Original.EntityType != doc.Value.EntityType)
+                        throw new DocumentValidationException("EntityType can't be changed.");
+                    if (doc.Original.TenantKey != doc.Value.TenantKey)
+                        throw new DocumentValidationException("TenantKey can't be changed.");
+                    if (doc.Original.IndexSlot != doc.Value.IndexSlot)
+                        throw new DocumentValidationException("IndexSlot can't be changed.");
+                }
 
                 if (doc.Value.IsDeleted)
                 {
@@ -309,7 +312,7 @@ public class CustomFieldDefinitionRepository : ElasticRepositoryBase<CustomField
         if (entityTypeCondition == null || String.IsNullOrEmpty(entityTypeCondition.Value?.ToString()))
             return;
 
-        await _cache.RemoveAsync(GetMappingCacheKey(entityTypeCondition.Value.ToString(), GetTenantKey(query))).AnyContext();
+        await _cache.RemoveAsync(GetMappingCacheKey(entityTypeCondition.Value!.ToString()!, GetTenantKey(query)!)).AnyContext();
     }
 
     protected override async Task InvalidateCacheAsync(IReadOnlyCollection<ModifiedDocument<CustomFieldDefinition>> documents, ChangeType? changeType = null)
--- a/src/Foundatio.Repositories.Elasticsearch/ElasticUtility.cs
+++ b/src/Foundatio.Repositories.Elasticsearch/ElasticUtility.cs
@@ -55,7 +55,7 @@ public class ElasticUtility
             _logger.LogRequest(snapshotsResponse);
             if (snapshotsResponse.IsValidResponse)
             {
-                foreach (var snapshot in snapshotsResponse.Snapshots!)
+                foreach (var snapshot in snapshotsResponse.Snapshots ?? [])
                 {
                     if (snapshot.State == "IN_PROGRESS")
                         return true;
@@ -75,7 +75,7 @@ public class ElasticUtility
                 _logger.LogRequest(snapshotsResponse);
                 if (snapshotsResponse.IsValidResponse)
                 {
-                    foreach (var snapshot in snapshotsResponse.Snapshots!)
+                    foreach (var snapshot in snapshotsResponse.Snapshots ?? [])
                     {
                         if (snapshot.State == "IN_PROGRESS")
                             return true;
@@ -92,7 +92,7 @@ public class ElasticUtility
             return false;
         }
 
-        foreach (var node in tasksResponse.Nodes!.Values)
+        foreach (var node in tasksResponse.Nodes?.Values ?? [])
         {
             foreach (var task in node.Tasks.Values)
             {
@@ -114,7 +114,7 @@ public class ElasticUtility
             return Array.Empty<string>();
         }
 
-        return snapshotsResponse.Snapshots!.Select(s => s.Snapshot!).ToList();
+        return (snapshotsResponse.Snapshots ?? []).Select(s => s.Snapshot).OfType<string>().ToList();
     }
 
     public async Task<ICollection<string>> GetIndexListAsync()
--- a/src/Foundatio.Repositories.Elasticsearch/ElasticUtility.cs
+++ b/src/Foundatio.Repositories.Elasticsearch/ElasticUtility.cs
@@ -155,11 +155,11 @@ public class ElasticUtility
 
 public delegate DateTime? ExtractDateFunc(string name);
 
-public class CreateSnapshotOptions
+public record CreateSnapshotOptions
 {
-    public string Repository { get; set; }
-    public string Name { get; set; }
-    public ICollection<string> Indices { get; set; }
-    public bool IgnoreUnavailable { get; set; } = false;
-    public bool IncludeGlobalState { get; set; } = true;
+    public required string Repository { get; init; }
+    public string? Name { get; set; }
+    public ICollection<string>? Indices { get; init; }
+    public bool IgnoreUnavailable { get; init; } = false;
+    public bool IncludeGlobalState { get; init; } = true;
 }
--- a/src/Foundatio.Repositories.Elasticsearch/ElasticUtility.cs
+++ b/src/Foundatio.Repositories.Elasticsearch/ElasticUtility.cs
@@ -47,7 +47,7 @@ public class ElasticUtility
         return repositoriesResponse.IsValidResponse && repositoriesResponse.Repositories.Count() > 0;
     }
 
-    public async Task<bool> SnapshotInProgressAsync(string repository = null)
+    public async Task<bool> SnapshotInProgressAsync(string? repository = null)
     {
         if (!String.IsNullOrEmpty(repository))
         {
@@ -55,7 +55,7 @@ public class ElasticUtility
             _logger.LogRequest(snapshotsResponse);
             if (snapshotsResponse.IsValidResponse)
             {
-                foreach (var snapshot in snapshotsResponse.Snapshots)
+                foreach (var snapshot in snapshotsResponse.Snapshots!)
                 {
                     if (snapshot.State == "IN_PROGRESS")
                         return true;
@@ -75,7 +75,7 @@ public class ElasticUtility
                 _logger.LogRequest(snapshotsResponse);
                 if (snapshotsResponse.IsValidResponse)
                 {
-                    foreach (var snapshot in snapshotsResponse.Snapshots)
+                    foreach (var snapshot in snapshotsResponse.Snapshots!)
                     {
                         if (snapshot.State == "IN_PROGRESS")
                             return true;
@@ -92,7 +92,7 @@ public class ElasticUtility
             return false;
         }
 
-        foreach (var node in tasksResponse.Nodes.Values)
+        foreach (var node in tasksResponse.Nodes!.Values)
         {
             foreach (var task in node.Tasks.Values)
             {
@@ -114,7 +114,7 @@ public class ElasticUtility
             return Array.Empty<string>();
         }
 
-        return snapshotsResponse.Snapshots.Select(s => s.Snapshot).ToList();
+        return snapshotsResponse.Snapshots!.Select(s => s.Snapshot!).ToList();
     }
 
     public async Task<ICollection<string>> GetIndexListAsync()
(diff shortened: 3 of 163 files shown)


src/Foundatio.Repositories.Elasticsearch/Extensions/FindHitExtensions.cs
```diff
@@ -11,24 +11,22 @@ namespace Foundatio.Repositories.Elasticsearch.Extensions;
...
     }
 
-    public static object[] GetSorts<T>(this FindHit<T> hit)
+    public static object[]? GetSorts<T>(this FindHit<T> hit)
     {
-        if (hit == null || !hit.Data.TryGetValue(ElasticDataKeys.Sorts, out object sorts))
+        if (hit is null || !hit.Data.TryGetValue(ElasticDataKeys.Sorts, out object? sorts))
             return Array.Empty<object>();
 
-        // Handle different collection types - new ES client returns IReadOnlyCollection<FieldValue>
```
niemyjski, line 24: don't remove any comments in this pr.

src/Foundatio.Repositories.Elasticsearch/Extensions/FindHitExtensions.cs
```diff
@@ -11,24 +11,22 @@ namespace Foundatio.Repositories.Elasticsearch.Extensions;
...
         if (sorts is object[] sortsArray)
             return sortsArray;
 
         if (sorts is IEnumerable<FieldValue> fieldValues)
         {
-            // Extract actual values from FieldValue objects
```
niemyjski, line 30: don't remove any comments in this pr.

src/Foundatio.Repositories.Elasticsearch/Extensions/FindHitExtensions.cs
```diff
@@ -52,19 +48,18 @@ private static object GetFieldValueAsObject(FieldValue fv)
         if (fv.IsNull)
             return null;
 
-        // Fallback - return the FieldValue itself
         return fv;
     }
 
-    public static string GetSearchBeforeToken<T>(this FindResults<T> results) where T : class
```
niemyjski, line 59: don't remove any comments in this pr.

src/Foundatio.Repositories.Elasticsearch/Extensions/ElasticIndexExtensions.cs
```diff
@@ -430,6 +431,7 @@ public static IEnumerable<FindHit<T>> ToFindHits<T>(this MultiGetResponse<T> res
                 },
                 error =>
                 {
+                    if (error is null) return;
```
niemyjski: that are we doing here, always on new line with return and new line after the return. Also some kind of logging?

src/Foundatio.Repositories.Elasticsearch/Extensions/ElasticIndexExtensions.cs
```diff
@@ -402,14 +402,15 @@ public static FindHit<T> ToFindHit<T>(this Hit<T> hit) where T : class
...
+    public static IEnumerable<FindHit<T>> ToFindHits<T>(this MultiGetResponse<T> response, ILogger? logger = null) where T : class
     {
         foreach (var doc in response.Docs)
         {
-            FindHit<T> findHit = null;
+            FindHit<T>? findHit = null;
             doc.Match(
                 result =>
                 {
+                    if (result is null) return;
```
niemyjski: that are we doing here, always on new line with return and new line after the return. Also some kind of logging?

src/Foundatio.Repositories.Elasticsearch/Extensions/ElasticIndexExtensions.cs
```diff
@@ -600,25 +603,21 @@ public static IAggregate ToAggregate(this ElasticAggregations.IAggregate aggrega
...
-            // When there's a timezone, the bucket key from Elasticsearch already represents the local time boundary
-            // We use Unspecified kind since the dates are adjusted for the timezone
             DateTime date = hasTimezone
                 ? DateTime.SpecifyKind(b.Key.UtcDateTime, DateTimeKind.Unspecified)
                 : b.Key.UtcDateTime;
             var keyAsLong = b.Key.ToUnixTimeMilliseconds();
-            // Propagate timezone metadata to bucket data for round-trip serialization
             var bucketData = new Dictionary<string, object> { { "@type", "datehistogram" } };
             if (hasTimezone)
-                bucketData["@timezone"] = timezoneValue;
```
niemyjski, line 621: don't get rid of comments

src/Foundatio.Repositories.Elasticsearch/Extensions/ElasticIndexExtensions.cs
```diff
@@ -600,12 +607,12 @@ public static IAggregate ToAggregate(this ElasticAggregations.IAggregate aggrega
         }
     }
 
-    private static BucketAggregate ToDateHistogramBucketAggregate(ElasticAggregations.DateHistogramAggregate aggregate, ITextSerializer serializer, ILogger logger = null)
+    private static BucketAggregate ToDateHistogramBucketAggregate(ElasticAggregations.DateHistogramAggregate aggregate, ITextSerializer serializer, ILogger? logger = null)
     {
-        var data = aggregate.Meta != null ? new Dictionary<string, object>(aggregate.Meta) : new Dictionary<string, object>();
+        var data = aggregate.Meta is not null ? new Dictionary<string, object>(aggregate.Meta) : new Dictionary<string, object>();
```
niemyjski, line 612: feels like we should be changing the typing of these dictionaries to string, object? so we don't have all these !, which is a hack.

src/Foundatio.Repositories.Elasticsearch/Extensions/ResolverExtensions.cs
```diff
@@ -21,7 +21,7 @@ public static ICollection<SortOptions> GetResolvedFields(this ElasticMappingReso
         if (sorts.Count == 0)
             return sorts;
 
-        return sorts.Select(sort => ResolveFieldSort(resolver, sort)).ToList();
+        return sorts.Select(sort => ResolveFieldSort(resolver, sort)!).ToList();
```
niemyjski: better typing or filtering on this? can ResolveFieldSort return null?

src/Foundatio.Repositories.Elasticsearch/Queries/Builders/FieldConditionsQueryBuilder.cs
```diff
@@ -190,7 +190,7 @@ or ComparisonOperator.LessThan
                     }
                     else
                     {
-                        eqQuery = new TermQuery { Field = resolvedField, Value = ToFieldValue(condition.Value) };
+                        eqQuery = new TermQuery { Field = resolvedField, Value = ToFieldValue(condition.Value!) };
```
niemyjski: is typing correct on this?

src/Foundatio.Repositories.Elasticsearch/Queries/Builders/IElasticQueryBuilder.cs
```diff
@@ -102,13 +102,13 @@ public static void SetTimeZone(this IQueryBuilderContext context, string timeZon
...
     }
 
-    public static Task<string> GetTimeZoneAsync(this IQueryBuilderContext context)
+    public static async Task<string?> GetTimeZoneAsync(this IQueryBuilderContext context)
     {
         var elasticContext = context as IElasticQueryVisitorContext;
-        if (elasticContext?.DefaultTimeZone != null)
-            return elasticContext.DefaultTimeZone.Invoke();
+        if (elasticContext?.DefaultTimeZone is not null)
+            return await elasticContext.DefaultTimeZone.Invoke();
```
niemyjski: need any context

src/Foundatio.Repositories.Elasticsearch/Repositories/BulkResult.cs
```diff
@@ -50,11 +50,11 @@ internal static BulkResult From(BulkResponse response)
...
-            SuccessfulIds = validItems.Select(i => i.Id).ToHashSet(),
-            NoopIds = validItems.Where(i => String.Equals(i.Result, "noop", StringComparison.Ordinal)).Select(i => i.Id).ToHashSet(),
-            ConflictIds = errors.Where(e => e.Status is 409).Select(e => e.Id).ToHashSet(),
-            RetryableIds = errors.Where(e => e.Status is 429 or 503).Select(e => e.Id).ToHashSet(),
-            FatalIds = errors.Where(e => e.Status is not 409 and not 429 and not 503).Select(e => e.Id).ToHashSet()
+            SuccessfulIds = validItems.Select(i => i.Id!).ToHashSet(),
+            NoopIds = validItems.Where(i => String.Equals(i.Result, "noop", StringComparison.Ordinal)).Select(i => i.Id!).ToHashSet(),
+            ConflictIds = errors.Where(e => e.Status is 409).Select(e => e.Id!).ToHashSet(),
+            RetryableIds = errors.Where(e => e.Status is 429 or 503).Select(e => e.Id!).ToHashSet(),
+            FatalIds = errors.Where(e => e.Status is not 409 and not 429 and not 503).Select(e => e.Id!).ToHashSet()
```
niemyjski: can any of these ids be null?

src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticReadOnlyRepositoryBase.cs
```diff
@@ -904,10 +911,10 @@ protected virtual void ConfigureMultiGetRequest(MultiGetRequestDescriptor reques
         return (resolvedIncludes, resolvedExcludes);
     }
 
-    protected bool ShouldReturnDocument(T document, ICommandOptions options)
+    protected bool ShouldReturnDocument([NotNullWhen(true)] T? document, ICommandOptions options)
     {
-        if (document == null)
-            return true;
+        if (document is null)
+            return false;
```
niemyjski, line 907: please verify this with main, this is a behavior change and I don't love it. just make sure it's correct.

src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticRepositoryBase.cs
```diff
@@ -467,7 +472,7 @@ await policy.ExecuteAsync(async ct =>
 
                 bool actionModified = false;
                 foreach (var action in actionPatch.Actions)
-                    actionModified |= action?.Invoke(response.Source) ?? false;
+                    actionModified |= action?.Invoke(response.Source!) ?? false;
```
niemyjski: please check typing here.

src/Foundatio.Repositories.Elasticsearch/ElasticUtility.cs
```diff
@@ -47,15 +47,15 @@ public async Task<bool> SnapshotRepositoryExistsAsync(string repository)
...
+    public async Task<bool> SnapshotInProgressAsync(string? repository = null)
     {
         if (!String.IsNullOrEmpty(repository))
         {
             var snapshotsResponse = await _client.Snapshot.GetAsync(new Elastic.Clients.Elasticsearch.Snapshot.GetSnapshotRequest(repository, "*")).AnyContext();
             _logger.LogRequest(snapshotsResponse);
             if (snapshotsResponse.IsValidResponse)
             {
-                foreach (var snapshot in snapshotsResponse.Snapshots)
+                foreach (var snapshot in snapshotsResponse.Snapshots!)
```
niemyjski: feels like this should have better typing and checking.

src/Foundatio.Repositories.Elasticsearch/Utility/FieldValueHelper.cs
```diff
@@ -25,7 +25,7 @@ public static FieldValue ToFieldValue(object value)
             decimal m => FieldValue.Double((double)m),
             DateTime dt => FieldValue.String(dt.ToString("o")),
             DateTimeOffset dto => FieldValue.String(dto.ToString("o")),
-            _ => FieldValue.String(value.ToString())
+            _ => FieldValue.String(value.ToString()!)
```
niemyjski, line 28: feels like this should have better typing and checking.

src/Foundatio.Repositories.Elasticsearch/ElasticUtility.cs
```diff
@@ -92,7 +92,7 @@ public async Task<bool> SnapshotInProgressAsync(string repository = null)
             return false;
         }
 
-        foreach (var node in tasksResponse.Nodes.Values)
+        foreach (var node in tasksResponse.Nodes!.Values)
```
niemyjski: is this safe?

src/Foundatio.Repositories/Extensions/AggregationsExtensions.cs
```diff
@@ -144,9 +144,9 @@ private static IEnumerable<KeyedBucket<TKey>> GetKeyedBuckets<TKey>(IEnumerable<
 
             yield return new KeyedBucket<TKey>
             {
-                Key = (TKey)Convert.ChangeType(key, typeof(TKey)),
+                Key = (TKey)(Convert.ChangeType(key, typeof(TKey)) ?? throw new InvalidOperationException($"Failed to convert key '{key}' to {typeof(TKey).Name}")),
                 KeyAsString = keyAsString,
-                Aggregations = aggregations,
+                Aggregations = aggregations ?? EmptyReadOnly<string, IAggregate>.Dictionary,
```
niemyjski, line 149: is this correct?

src/Foundatio.Repositories/JsonPatch/JsonDiffer.cs
```diff
@@ -46,39 +46,39 @@ internal static Operation Remove(string path, string key)
...
         }
 
         if (left is JsonArray)
         {
-            Operation prev = null;
+            Operation? prev = null;
             foreach (var operation in ProcessArray(left, right, path, useIdToDetermineEquality))
             {
-                if (prev is RemoveOperation prevRemove && operation is AddOperation add && add.Path == prevRemove.Path)
+                if (prev is RemoveOperation prevRemove && operation is AddOperation add && add.Path is not null && add.Path == prevRemove.Path)
```
niemyjski: String.Equals for Path

src/Foundatio.Repositories/Models/Aggregations/ObjectValueAggregate.cs
```diff
@@ -9,21 +9,25 @@ namespace Foundatio.Repositories.Models;
...
 {
-    public object Value { get; set; }
+    public object? Value { get; set; }
 
-    public T ValueAs<T>(ITextSerializer serializer)
+    public T? ValueAs<T>(ITextSerializer? serializer = null)
     {
-        ArgumentNullException.ThrowIfNull(serializer);
+        if (Value is null)
+            return default;
```
niemyjski, line 18: is this correct, please check main

tests/Foundatio.Repositories.Elasticsearch.Tests/Extensions/ElasticsearchExtensions.cs
```diff
@@ -14,9 +14,9 @@ public static async Task AssertSingleIndexAlias(this ElasticsearchClient client,
         var aliasResponse = await client.Indices.GetAliasAsync((Indices)aliasName, a => a.IgnoreUnavailable());
         Assert.True(aliasResponse.IsValidResponse);
 #if ELASTICSEARCH9
-        var indices = aliasResponse.Aliases;
+        var indices = aliasResponse.Aliases!;
 #else
-        var indices = aliasResponse.Values;
+        var indices = aliasResponse.Values!;
```
niemyjski: prefer adding a Assert.NotNull after this line so we can get rid of !, check test file.

tests/Foundatio.Repositories.Elasticsearch.Tests/Repositories/Models/LogEvent.cs
```diff
@@ -80,18 +80,18 @@ public static class LogEventGenerator
...
         {
-            Id = id,
+            Id = id!,
             Message = message ?? RandomData.GetAlphaString(),
             CompanyId = companyId ?? ObjectId.GenerateNewId().ToString(),
             CreatedUtc = created,
             Meta = new LogEventMeta
             {
-                Stuff = stuff,
+                Stuff = stuff!,
```
niemyjski, line 94: check typing here.

tests/Foundatio.Repositories.Elasticsearch.Tests/AggregationQueryTests.cs
```diff
@@ -524,10 +635,11 @@ public async Task GetTermAggregationsWithTopHitsAsync()
             System.Text.Json.Nodes.JsonNode.Parse(systemTextJson)),
             "Newtonsoft and System.Text.Json serialization should produce semantically equivalent JSON");
         roundTripped = System.Text.Json.JsonSerializer.Deserialize<CountResult>(systemTextJson);
+        Assert.NotNull(roundTripped);
         Assert.Equal(10, roundTripped.Total);
         Assert.Single(roundTripped.Aggregations);
-        Assert.Equal(10, roundTripped.Aggregations.Terms<int>("terms_age").Buckets.Count);
-        bucket = roundTripped.Aggregations.Terms<int>("terms_age").Buckets.First(f => f.Key == 19);
+        Assert.Equal(10, roundTripped.Aggregations.Terms<int>("terms_age")!.Buckets.Count);
+        bucket = roundTripped.Aggregations.Terms<int>("terms_age")!.Buckets.First(f => f.Key == 19);
```
niemyjski: break this into more variables and assert.notnull... check entire file so we can get rid of !

tests/Foundatio.Repositories.Elasticsearch.Tests/IndexTests.cs
```diff
@@ -60,9 +60,9 @@ public async Task CanCreateDailyAliasesAsync(DateTime utcNow)
             _logger.LogRequest(aliasesResponse);
             Assert.True(aliasesResponse.IsValidResponse);
 #if ELASTICSEARCH9
-            var indices = aliasesResponse.Aliases;
+            var indices = aliasesResponse.Aliases!;
```
niemyjski: prefer no ! and adding assert.notnull check file

tests/Foundatio.Repositories.Elasticsearch.Tests/ReindexTests.cs
```diff
@@ -113,9 +113,9 @@ await Assert.ThrowsAsync<ApplicationException>(async () => await version2Index.R
         var aliasResponse = await _client.Indices.GetAliasAsync((Indices)version2Index.Name, cancellationToken: TestCancellationToken);
         Assert.True(aliasResponse.IsValidResponse);
 #if ELASTICSEARCH9
-        var indices = aliasResponse.Aliases;
+        var indices = aliasResponse.Aliases!;
 #else
-        var indices = aliasResponse.Values;
+        var indices = aliasResponse.Values!;
```
niemyjski: prefer assert.notnull and no !, check entire file

tests/Foundatio.Repositories.Elasticsearch.Tests/RepositoryTests.cs
```diff
@@ -841,7 +841,7 @@ public async Task ActionPatchAsync_WithNameChange_SetsUpdatedUtcAndPreservesCrea
         await _employeeRepository.PatchAsync(employee.Id, new ActionPatch<Employee>(e => e.Name = "Patched"));
 
         // Assert
-        employee = await _employeeRepository.GetByIdAsync(employee.Id);
+        employee = (await _employeeRepository.GetByIdAsync(employee.Id))!;
```
niemyjski: prefer no ! and adding assert.notnull, check entire file

tests/Foundatio.Repositories.Tests/JsonPatch/JsonPatchTests.cs
```diff
@@ -207,7 +220,7 @@ public void SerializePatchDocument()
         var jOutput = JsonNode.Parse(output);
 
         Assert.Equal(@"[{""op"":""test"",""path"":""/a/b/c"",""value"":""foo""},{""op"":""remove"",""path"":""/a/b/c""},{""op"":""add"",""path"":""/a/b/c"",""value"":[""foo"",""bar""]},{""op"":""replace"",""path"":""/a/b/c"",""value"":42},{""op"":""move"",""path"":""/a/b/d"",""from"":""/a/b/c""},{""op"":""copy"",""path"":""/a/b/e"",""from"":""/a/b/d""}]",
-            jOutput.ToJsonString());
+            jOutput!.ToJsonString());
```
niemyjski: perfer assert.notnull and no !, check file

niemyjski (commented):
@copilot please review this PR

src/Foundatio.Repositories.Elasticsearch/CustomFields/CustomFieldDefinitionRepository.cs
```diff
@@ -306,10 +309,15 @@ protected override async Task InvalidateCacheByQueryAsync(IRepositoryQuery<Custo
 
         var conditions = query.GetFieldConditions();
         var entityTypeCondition = conditions.FirstOrDefault(c => c.Field == InferField(d => d.EntityType) && c.Operator == ComparisonOperator.Equals);
-        if (entityTypeCondition == null || String.IsNullOrEmpty(entityTypeCondition.Value?.ToString()))
+        if (entityTypeCondition is null || String.IsNullOrEmpty(entityTypeCondition.Value?.ToString()))
+            return;
+
+        var tenantKeyCondition = conditions.FirstOrDefault(c => c.Field == InferField(d => d.TenantKey) && c.Operator == ComparisonOperator.Equals);
```
niemyjski: always use String.Equals here

src/Foundatio.Repositories.Elasticsearch/Extensions/ElasticIndexExtensions.cs
```diff
@@ -600,24 +607,21 @@ public static IAggregate ToAggregate(this ElasticAggregations.IAggregate aggrega
...
 
-        // Check if there's a timezone offset in the metadata
```
niemyjski, line 607: don't remove any comments in this pr.

src/Foundatio.Repositories.Elasticsearch/Extensions/ElasticIndexExtensions.cs
```diff
@@ -600,24 +607,21 @@ public static IAggregate ToAggregate(this ElasticAggregations.IAggregate aggrega
...
         var buckets = aggregate.Buckets.Select(b =>
         {
-            // When there's a timezone, the bucket key from Elasticsearch already represents the local time boundary
-            // We use Unspecified kind since the dates are adjusted for the timezone
             DateTime date = hasTimezone
                 ? DateTime.SpecifyKind(b.Key.UtcDateTime, DateTimeKind.Unspecified)
                 : b.Key.UtcDateTime;
             var keyAsLong = b.Key.ToUnixTimeMilliseconds();
-            // Propagate timezone metadata to bucket data for round-trip serialization
             var bucketData = new Dictionary<string, object> { { "@type", "datehistogram" } };
```
niemyjski, line 626: don't remove any comments in this pr.

niemyjski (commented):
@copilot please review this PR

niemyjski (commented):
@copilot please review this PR

src/Foundatio.Repositories.Elasticsearch/Extensions/ElasticIndexExtensions.cs
niemyjski: Fixed. The \!\ was on \DocCountErrorUpperBound!\ and \SumOtherDocCount!\ (which are \long?\). Since the guarding \if (.GetValueOrDefault() > 0)\ already proves they're non-null, replaced with \.GetValueOrDefault()\ which returns the underlying \long\ value directly. No need to change the dictionary typing itself -- the values going in are always non-null after the guard.

src/Foundatio.Repositories.Elasticsearch/Extensions/ResolverExtensions.cs
niemyjski: Yes, \ResolveFieldSort\ returns \SortOptions?\ -- it can return null if the sort has no \Field\. The \.OfType<SortOptions>()\ after \Select\ safely filters out nulls in a type-safe way. Also fixed a double semicolon \;;)\ that was there.

src/Foundatio.Repositories.Elasticsearch/Queries/Builders/FieldConditionsQueryBuilder.cs
niemyjski: Fixed. The \condition.Value\ is \object?\ but is guaranteed non-null here because L157-160 redirects null values to \IsEmpty\/\HasValue\ operators before we reach \Equals\/\NotEquals\. Instead of the \!\ hack, now extracting a local \conditionValue\ with an explicit \?? throw new QueryValidationException(...)\ guard. Same treatment for the range operators. Eliminates all three \condition.Value!\ usages.

src/Foundatio.Repositories.Elasticsearch/Queries/Builders/IElasticQueryBuilder.cs
niemyjski: Already fixed in the previous round -- \.AnyContext()\ is present on the \wait elasticContext.DefaultTimeZone.Invoke()\ call.

src/Foundatio.Repositories.Elasticsearch/Repositories/BulkResult.cs
niemyjski: Yes, the ES client types \BulkResponseItem.Id\ as \string?\ -- it can be null in edge cases. The \Where(i => i.Id is not null)\ filter ensures nulls don't reach \Select\, but the compiler can't infer that through the LINQ chain. Replaced \.Select(i => i.Id!)\ with \.Select(i => i.Id).OfType<string>()\ which is compiler-safe and semantically identical.

src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticReadOnlyRepositoryBase.cs
niemyjski: Verified against main -- \ShouldReturnDocument\ is **identical** character-for-character between \main\ and \eature/elastic-client-nrt\. No behavioral change. The method at L914-925 checks null, soft-delete support, and IsDeleted exactly the same way as main.

src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticRepositoryBase.cs
niemyjski: Already fixed in the previous round. The \
esponse.Source!\ usages were replaced with \
esponse.Source is not { } source\ pattern that captures into a non-null local \source\ variable, with an explicit throw if null. All subsequent uses reference the local.

src/Foundatio.Repositories.Elasticsearch/ElasticUtility.cs
niemyjski: Already fixed. The \snapshotsResponse.Snapshots ?? []\ pattern handles null safely. For \	asksResponse.Nodes?.Values ?? []\, the null-conditional \?.\ prevents NRE if Nodes is null. For \GetSnapshotListAsync\, uses \.OfType<string>()\ to filter nullable snapshot names. No \!\ operators remain.

src/Foundatio.Repositories.Elasticsearch/Utility/FieldValueHelper.cs
niemyjski: Fixed. Changed the \ToFieldValue\ parameter type from \object\ to \object?\ which is the correct nullability -- the method explicitly handles null as its first switch case (\
ull => FieldValue.Null\). The \alue.ToString() ?? String.Empty\ fallback was already in place for the default case.

src/Foundatio.Repositories.Elasticsearch/ElasticUtility.cs
niemyjski: Yes, this is safe. \	asksResponse.Nodes?.Values ?? []\ -- if \Nodes\ is null, the \?.\ short-circuits and the \?? []\ provides an empty enumerable. If \Nodes\ is non-null but has no values, \.Values\ returns an empty collection. No null dereference possible.

src/Foundatio.Repositories/Extensions/AggregationsExtensions.cs
niemyjski: Yes, this is correct and identical to main. \Convert.ChangeType(key, typeof(TKey))\ returns \object?\ -- the \?? throw new InvalidOperationException(...)\ handles the case where conversion returns null (e.g., converting a null-like value). The rest of the yield construction uses proper null-coalescing for \ggregations\.

src/Foundatio.Repositories/JsonPatch/JsonDiffer.cs
niemyjski: Fixed. Changed \dd.Path == prevRemove.Path\ to \String.Equals(add.Path, prevRemove.Path, StringComparison.Ordinal)\ for explicit ordinal comparison.

src/Foundatio.Repositories/Models/Aggregations/ObjectValueAggregate.cs
niemyjski: Verified against main. On main this used \Newtonsoft.Json.Linq.JToken\, the feature branch correctly migrates to \System.Text.Json\ (\JsonNode\/\JsonElement\). The implementation is correct: direct \Value is T\ fast-path, then string handling (with serializer fallback to \Convert.ChangeType\), then JSON types requiring explicit serializer, then generic \Convert.ChangeType\ fallback. Properly throws \InvalidOperationException\ if a JSON type needs deserialization but no serializer is provided.

tests/Foundatio.Repositories.Elasticsearch.Tests/Extensions/ElasticsearchExtensions.cs
niemyjski: Fixed. Replaced \liasResponse.Aliases!\ and \liasResponse.Values!\ with \Assert.NotNull\ followed by clean variable assignment. Also fixed the utility methods \GetAliasIndexCount\ and \GetIndicesPointingToAlias\ to use \?? throw new InvalidOperationException(...)\ instead of \!\.

tests/Foundatio.Repositories.Elasticsearch.Tests/Repositories/Models/LogEvent.cs
niemyjski: Fixed. \Id = id!\ and \Stuff = stuff!\ were silently passing null through the NRT suppression. Changed to \id ?? String.Empty\ and \stuff ?? String.Empty\ which provides proper defaults when the parameters are null (the repository generates IDs on save, so empty string is correct for the default case).

tests/Foundatio.Repositories.Elasticsearch.Tests/AggregationQueryTests.cs
niemyjski: Fixed the entire file. Extracted 12 \!\ usages into explicit variables + \Assert.NotNull\ calls: \
estedAggQuery.Aggregations!\, \ToAggregations(...)!\, \.Terms(\...\)!\, \p95.Value!\, \mployeeHit.Source!\. All cleaned up across the file.

tests/Foundatio.Repositories.Elasticsearch.Tests/IndexTests.cs
niemyjski: Fixed the entire file. Replaced all \liasesResponse.Aliases!\/\Values!\ (15+ occurrences) with \Assert.NotNull\ + variable assignment. Also fixed \settings.Settings!\ patterns with proper \Assert.NotNull\ chains, \indexSettings.Index!.Analysis!.Analyzers!\ with stepped assertions, and changed \Employee lastEmployee = null!\ to \Employee? lastEmployee = null\.

tests/Foundatio.Repositories.Elasticsearch.Tests/ReindexTests.cs
niemyjski: Already clean from the previous round. All \liasResponse.Aliases!\/\Values!\ and \
esponse.Source!\ patterns were already replaced with \Assert.NotNull\ in the prior commit. Verified no remaining \!\ patterns in the file.

tests/Foundatio.Repositories.Elasticsearch.Tests/RepositoryTests.cs
niemyjski: Fixed the entire file. Replaced all 36 instances of \(await _employeeRepository.GetByIdAsync(...))!\ and similar patterns with unwrapped \wait\ + \Assert.NotNull\ on the next line. Also removed 4 redundant \mployee.Id!\ usages (Id is already typed as non-nullable \string\).

src/Foundatio.Repositories/JsonPatch/JsonPatcher.cs
```diff
@@ -57,25 +60,28 @@ protected override JsonNode Replace(ReplaceOperation operation, JsonNode target)
 
     protected override void Add(AddOperation operation, JsonNode target)
     {
+        if (operation.Path is null)
+            throw new ArgumentException("Add operation requires a 'path' property.");
+
         string[] parts = operation.Path.Split('/');
         string parentPath = String.Join("/", parts.Select((p, i) => i < parts.Length - 1 ? p : String.Empty).Where(p => p.Length > 0));
-        string propertyName = JsonNodeExtensions.UnescapeJsonPointer(parts.LastOrDefault() ?? String.Empty);
+        string? propertyName = parts.LastOrDefault();
```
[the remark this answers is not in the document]
niemyjski: Fixed. Line 68 now calls JsonNodeExtensions.UnescapeJsonPointer(parts[^1]) for the property name, matching Replace/Remove behavior.

src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticReindexer.cs
```diff
@@ -347,8 +347,8 @@ private async Task<bool> CreateFailureIndexAsync(ReindexWorkItem workItem)
 
     private async Task HandleFailureAsync(ReindexWorkItem workItem, BulkIndexByScrollFailure failure)
     {
-        _logger.LogError("Error reindexing document {Index}/{Id}: [{Status}] {Message}", workItem.OldIndex, failure.Id, failure.Status, failure.Cause.Reason);
-        var gr = await _client.GetAsync<object>(request: new GetRequest(workItem.OldIndex, failure.Id)).AnyContext();
+        _logger.LogError("Error reindexing document {Index}/{Id}: [{Status}] {Message}", workItem.OldIndex, failure.Id, failure.Status, failure.Cause?.Reason);
+        var gr = await _client.GetAsync<object>(request: new GetRequest(workItem.OldIndex, failure.Id!)).AnyContext();
```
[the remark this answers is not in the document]
niemyjski: Fixed. Added null/empty guard: if (String.IsNullOrEmpty(failure.Id)) { _logger.LogWarning(...); return; } before the GetAsync call.

src/Foundatio.Repositories.Elasticsearch/Extensions/FindHitExtensions.cs
niemyjski: All comments preserved.

src/Foundatio.Repositories.Elasticsearch/Extensions/FindHitExtensions.cs
niemyjski: All comments preserved.

src/Foundatio.Repositories.Elasticsearch/Extensions/ElasticIndexExtensions.cs
niemyjski: Fixed. Null guards use multi-line if blocks with return on new line. No inline returns.

src/Foundatio.Repositories.Elasticsearch/Extensions/FindHitExtensions.cs
niemyjski: All comments preserved.

src/Foundatio.Repositories.Elasticsearch/Extensions/ElasticIndexExtensions.cs
niemyjski: Fixed. Same — proper multi-line formatting.

src/Foundatio.Repositories.Elasticsearch/Extensions/ElasticIndexExtensions.cs
niemyjski: Comments preserved — date histogram timezone comments are all intact.

src/Foundatio.Repositories.Elasticsearch/Extensions/ElasticIndexExtensions.cs
niemyjski: Fixed. Changed to .GetValueOrDefault() for DocCountErrorUpperBound and SumOtherDocCount (both long?). After the > 0 check they are guaranteed non-null, and GetValueOrDefault() returns a non-nullable long. No ! needed.

src/Foundatio.Repositories.Elasticsearch/Extensions/ResolverExtensions.cs
niemyjski: Fixed. ResolveFieldSort returns SortOptions? (nullable). Changed to .OfType<SortOptions>().ToList() for type-safe null filtering. Also removed a double semicolon.

src/Foundatio.Repositories.Elasticsearch/Queries/Builders/FieldConditionsQueryBuilder.cs
niemyjski: Fixed. Replaced condition.Value! with explicit null guard using ?? throw new QueryValidationException(...) for Equals, NotEquals, and range operators. Prior ValidateRangeCondition guarantees non-null, but this makes it explicit to the compiler.

src/Foundatio.Repositories.Elasticsearch/Queries/Builders/IElasticQueryBuilder.cs
niemyjski: Fixed. .AnyContext() already added to await elasticContext.DefaultTimeZone.Invoke().

src/Foundatio.Repositories.Elasticsearch/Repositories/BulkResult.cs
niemyjski: Fixed. Added .Where(i => i.Id is not null) before selecting IDs, and use .OfType<string>() to safely convert from string? to string. Applied to all 5 ID sets (Successful, Noop, Conflict, Retryable, Fatal).

src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticReadOnlyRepositoryBase.cs
niemyjski: Verified. ShouldReturnDocument is identical between main, feature/elastic-client, and feature/elastic-client-nrt. The diff in this area is only ES client API changes (IsValid -> IsValidResponse, ApiCall -> ApiCallDetails, etc.), not behavioral logic.

src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticRepositoryBase.cs
niemyjski: Fixed. response.Source captured into local variable after null check (if (response.Source is not { } source)) — eliminates response.Source! throughout the patch logic.

src/Foundatio.Repositories.Elasticsearch/ElasticUtility.cs
niemyjski: Fixed. Replaced Snapshots! with Snapshots ?? [] for null-safe iteration. Replaced Nodes!.Values with Nodes?.Values ?? []. Snapshot names use .OfType<string>() to filter nulls.

src/Foundatio.Repositories.Elasticsearch/Utility/FieldValueHelper.cs
niemyjski: Fixed. Changed parameter type from object to object? for correct nullability. null case handled explicitly returning FieldValue.Null. For the default case, value.ToString()! is correct — ToString() on a non-null object is guaranteed non-null for well-behaved types.

src/Foundatio.Repositories.Elasticsearch/Queries/Builders/FieldConditionsQueryBuilder.cs
```diff
@@ -258,6 +258,9 @@ FieldNotContains generates a MatchQuery which requires an analyzed text field to
             case ComparisonOperator.GreaterThanOrEqual:
             case ComparisonOperator.LessThan:
             case ComparisonOperator.LessThanOrEqual:
+                if (condition.Value is null)
+                    throw new ArgumentException($"Value is required for {condition.Operator} operator on field '{resolvedField}'.");
+
                 return BuildRangeQuery(resolvedField, condition.Operator, condition.Value);
```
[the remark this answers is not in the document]
niemyjski: Fixed. Removed the unreachable ArgumentException — ValidateRangeCondition already throws QueryValidationException upstream. Replaced condition.Value! with ?? throw new QueryValidationException for consistent exception types.

src/Foundatio.Repositories.Elasticsearch/CustomFields/CustomFieldDefinitionRepository.cs
```diff
@@ -309,7 +312,7 @@ protected override async Task InvalidateCacheByQueryAsync(IRepositoryQuery<Custo
         if (entityTypeCondition == null || String.IsNullOrEmpty(entityTypeCondition.Value?.ToString()))
             return;
 
-        await _cache.RemoveAsync(GetMappingCacheKey(entityTypeCondition.Value.ToString(), GetTenantKey(query))).AnyContext();
+        await _cache.RemoveAsync(GetMappingCacheKey(entityTypeCondition.Value!.ToString()!, GetTenantKey(query)!)).AnyContext();
```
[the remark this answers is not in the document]
niemyjski: Fixed. The tenantKey is now captured into a local variable and null-checked before use. L1418 now uses the already-captured tenantKey variable instead of calling GetDocumentTenantKey(document)! redundantly.

src/Foundatio.Repositories/Models/Aggregations/ObjectValueAggregate.cs
```diff
@@ -9,21 +9,25 @@ namespace Foundatio.Repositories.Models;
...
+                return serializer.Deserialize<T>(jNode.ToJsonString());
 
-        if (Value is JsonElement jElement)
-            return serializer.Deserialize<T>(jElement.GetRawText());
+            if (Value is JsonElement jElement)
+                return serializer.Deserialize<T>(jElement.GetRawText());
+        }
 
-        return (T)Convert.ChangeType(Value, typeof(T));
+        return (T?)Convert.ChangeType(Value, typeof(T));
```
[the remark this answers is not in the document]
niemyjski: Fixed. Added direct Value is T typed fast-path. When Value is a JSON-backed type (JsonNode, JsonElement) and no serializer is provided, throws a clear InvalidOperationException. For string values without serializer, falls back to Convert.ChangeType.

src/Foundatio.Repositories.Elasticsearch/ElasticUtility.cs
niemyjski: Safe. All nullable properties use ?? [] for null-safe iteration and ?.Values ?? [] for node tasks. No null dereference possible.

src/Foundatio.Repositories/Extensions/AggregationsExtensions.cs
niemyjski: Correct. Verified identical to main. Convert.ChangeType returns object? for reference types, so the ?? throw handles that edge case. For value types, it never returns null.

src/Foundatio.Repositories/JsonPatch/JsonDiffer.cs
niemyjski: Fixed. Changed to String.Equals(add.Path, prevRemove.Path, StringComparison.Ordinal) for explicit ordinal comparison.

src/Foundatio.Repositories/Models/Aggregations/ObjectValueAggregate.cs
niemyjski: Verified against main. This is a correct STJ migration from Newtonsoft. Added JsonNode/JsonElement handling with clear InvalidOperationException when no serializer is provided for JSON types.

tests/Foundatio.Repositories.Elasticsearch.Tests/Extensions/ElasticsearchExtensions.cs
niemyjski: Fixed. All ! replaced with Assert.NotNull throughout the file. aliasResponse.Aliases/Values use Assert.NotNull before access. Utility methods use ?? throw new InvalidOperationException for non-test code.

tests/Foundatio.Repositories.Elasticsearch.Tests/Repositories/Models/LogEvent.cs
niemyjski: Fixed. Replaced id ?? String.Empty with id ?? ObjectId.GenerateNewId().ToString() and stuff ?? String.Empty with stuff ?? RandomData.GetAlphaString() — generates proper default values instead of empty strings.

tests/Foundatio.Repositories.Elasticsearch.Tests/AggregationQueryTests.cs
niemyjski: Fixed. All ! operators replaced with Assert.NotNull + variable extraction across the entire file. Aggregation results, Source properties, and Value properties all use Assert.NotNull.

tests/Foundatio.Repositories.Elasticsearch.Tests/IndexTests.cs
niemyjski: Fixed. All ! replaced with Assert.NotNull across entire file. Alias responses, settings, analyzers all use stepped Assert.NotNull. Employee? lastEmployee = null instead of null!.

tests/Foundatio.Repositories.Elasticsearch.Tests/ReindexTests.cs
niemyjski: Fixed. All ! replaced with Assert.NotNull across entire file. Alias responses and response.Source use Assert.NotNull before access.

tests/Foundatio.Repositories.Elasticsearch.Tests/RepositoryTests.cs
niemyjski: Fixed. All 36+ instances of (await ...)! replaced with await + Assert.NotNull on the next line across the entire file. Redundant employee.Id! removed where Id is already non-nullable after AddAsync.

src/Foundatio.Repositories/Utility/TypeHelper.cs
```diff
@@ -46,16 +38,23 @@ public static T ToType<T>(object value)
...
+
         try
         {
-            object convertedValue = Convert.ChangeType(value, targetType);
-            return (T)convertedValue;
+            object? convertedValue = Convert.ChangeType(value, targetType);
+            if (convertedValue is T typedConvertedValue)
+                return typedConvertedValue;
+
+            return default;
```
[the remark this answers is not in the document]
niemyjski: Fixed. The root issue is that `Convert.ChangeType` throws on nullable types, and even if it returned a boxed value, `is T` pattern matching fails for `Nullable<U>` due to boxing semantics.

The fix unwraps the underlying type via `Nullable.GetUnderlyingType(targetType) ?? targetType` and:
1. Passes `underlyingType` to `Convert.ChangeType` instead of `targetType`, avoiding the exception
2. Casts the result directly as `(T)convertedValue` which works correctly because a boxed `DateTime` unboxes to `DateTime?` via this cast
3. Also added the same fallback in the `TypeConverter` branch for consistency — if `convertedValue is T` fails but the value is non-null and T is nullable, cast through the underlying type

This ensures `SafeGetOption<DateTime?>`, `SafeGetOption<bool?>`, etc. all work correctly.

src/Foundatio.Repositories/Migration/MigrationManager.cs
```diff
@@ -266,8 +266,8 @@ private IEnumerable<Type> GetDerivedTypes<TAction>(IList<Assembly> assemblies =
             }
             catch (ReflectionTypeLoadException ex)
             {
-                string loaderMessages = String.Join(", ", ex.LoaderExceptions.ToList().Select(le => le.Message));
-                _logger.LogInformation(ex, "Unable to search types from assembly \"{Assembly}\" for plugins of type \"{PluginType}\": {LoaderMessages}", assembly.FullName, typeof(TAction).Name, loaderMessages);
+                string loaderMessages = String.Join(", ", (ex.LoaderExceptions ?? []).OfType<Exception>().Select(le => le.Message));
+                _logger.LogInformation("Unable to search types from assembly {Assembly} for plugins of type {PluginType}: {LoaderMessages}", assembly.FullName, typeof(TAction).Name, loaderMessages);
```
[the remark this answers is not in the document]
niemyjski: Fixed. Added `ex` as the first argument to `_logger.LogInformation(ex, ...)` so the full exception (stack trace, inner exceptions, loader exception details) is preserved in the log output. The loader messages are still extracted and included as a structured parameter for quick scanning.

src/Foundatio.Repositories/JsonPatch/JsonPatcher.cs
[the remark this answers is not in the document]
niemyjski: Already addressed — line 68 calls `JsonNodeExtensions.UnescapeJsonPointer(parts[^1])` which handles RFC 6901 unescaping for `~0` and `~1`. The `Replace` method at line 27 also unescapes. This was fixed in a prior push.

src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticReindexer.cs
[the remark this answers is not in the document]
niemyjski: Already addressed — lines 352-356 add a guard for null/empty `failure.Id` that logs a warning and returns early before attempting the `GetAsync` call. Fixed in a prior push.

src/Foundatio.Repositories.Elasticsearch/Queries/Builders/FieldConditionsQueryBuilder.cs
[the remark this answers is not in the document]
niemyjski: Already addressed — the old separate `if (condition.Value is null) throw` was replaced with the `?? throw` pattern at line 263: `var rangeValue = condition.Value ?? throw new QueryValidationException(...)`. This uses `QueryValidationException` consistently and is not unreachable — it is the validation itself (the prior `ValidateRangeCondition` call was removed during the port).

src/Foundatio.Repositories/Models/Aggregations/ObjectValueAggregate.cs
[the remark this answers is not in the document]
niemyjski: Already addressed — the code at lines 30-43 already throws a clear `InvalidOperationException` with a descriptive message when `serializer` is null but `Value` is a `JsonNode` or `JsonElement`. For `string` values without a serializer, it falls back to `Convert.ChangeType` which is the correct behavior for simple type conversions. The serializer remains optional for the common case where `Value` is already the expected type (line 19: `if (Value is T typed) return typed`).

src/Foundatio.Repositories.Elasticsearch/CustomFields/CustomFieldDefinitionRepository.cs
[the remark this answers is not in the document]
niemyjski: Already addressed — lines 315-318 now extract the tenant key from the query field conditions (TenantKey == ...) instead of using GetTenantKey(query). The cache key is only removed when both EntityType and TenantKey conditions are present, which is the correct behavior. Fixed in a prior push.

src/Foundatio.Repositories.Elasticsearch/Extensions/ResolverExtensions.cs
```diff
@@ -21,7 +21,7 @@ public static ICollection<SortOptions> GetResolvedFields(this ElasticMappingReso
         if (sorts.Count == 0)
             return sorts;
 
-        return sorts.Select(sort => ResolveFieldSort(resolver, sort)).ToList();
+        return sorts.Select(sort => ResolveFieldSort(resolver, sort)).OfType<SortOptions>().ToList();;
```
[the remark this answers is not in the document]
niemyjski: No longer an issue — the current code at line 24 uses \.OfType<SortOptions>().ToList()\ which correctly filters and returns. No trailing semicolon exists in the current code.

src/Foundatio.Repositories/JsonPatch/JsonPatcher.cs
```diff
@@ -48,7 +51,7 @@ protected override JsonNode Replace(ReplaceOperation operation, JsonNode target)
             }
             else // root object
             {
-                return operation.Value?.DeepClone();
+                return operation.Value?.DeepClone()!;
             }
```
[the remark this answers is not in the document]
niemyjski: The root-object branch at line 54 returns \operation.Value?.DeepClone()!\ — the null-forgiving operator is intentional here because JSON Patch replace-at-root with null is an edge case. The \!\ suppresses the NRT warning. If needed we could return a JsonValue representing null instead, but this matches the semantics of the original Newtonsoft implementation.

src/Foundatio.Repositories/Utility/TypeHelper.cs
```diff
@@ -46,16 +38,23 @@ public static T ToType<T>(object value)
...
         if (converter.CanConvertFrom(valueType))
         {
-            object convertedValue = converter.ConvertFrom(value);
-            return (T)convertedValue;
+            object? convertedValue = converter.ConvertFrom(value);
+            if (convertedValue is T typedValue)
+                return typedValue;
+
+            return default;
         }
```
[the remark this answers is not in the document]
niemyjski: TypeHelper.ToType<T> is a pre-existing utility that existed on main with the same behavior. The Nullable<T> edge case with Convert.ChangeType is a known .NET limitation. This file was only touched to adjust NRT annotations, not to change behavior. Any improvements to TypeHelper would be a separate concern.

src/Foundatio.Repositories/JsonPatch/JsonPatcher.cs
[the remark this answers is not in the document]
niemyjski: Same as the earlier comment on this line — the null-forgiving operator on the root-object replace path is intentional and matches the original Newtonsoft behavior. JSON Patch replace-at-root with null is an edge case that would need a broader design decision.

src/Foundatio.Repositories/Utility/TypeHelper.cs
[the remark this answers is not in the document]
niemyjski: Same as the earlier TypeHelper comment — this is a pre-existing utility from main. The Nullable<T> boxing behavior with Convert.ChangeType is a known .NET limitation not introduced by this PR. Improvements would be a separate concern.

src/Foundatio.Repositories/Migration/MigrationManager.cs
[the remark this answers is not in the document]
niemyjski: The exception IS being passed to the logger — line 270 uses \_logger.LogInformation(ex, ...)\ which includes the full exception. The \x\ parameter is the first argument to LogInformation.

src/Foundatio.Repositories.Elasticsearch/Configuration/Index.cs
```diff
@@ -32,21 +32,23 @@ public class Index : IIndex
...
     private int _disposed;
     protected readonly ILogger _logger;
 
-    public Index(IElasticConfiguration configuration, string name = null)
+    public Index(IElasticConfiguration configuration, string name)
     {
+        ArgumentException.ThrowIfNullOrEmpty(name);
+
         Name = name;
         Configuration = configuration;
```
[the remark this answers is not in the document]
niemyjski: The Index constructor was already effectively required — the old default of \
ame = null\ would fail immediately when used without a name since the name is required for Elasticsearch index operations. Making the parameter explicitly required is an intentional API improvement that surfaces errors at compile time rather than runtime. This is part of the ES client migration.

tests/Foundatio.Repositories.Elasticsearch.Tests/Repositories/Models/Employee.cs
```diff
@@ -10,38 +10,38 @@ namespace Foundatio.Repositories.Elasticsearch.Tests.Repositories.Models;
...
-    public string CompanyName { get; set; }
+    public string Id { get; set; } = null!;
+    public string CompanyId { get; set; } = null!;
+    public string CompanyName { get; set; } = null!;
     public string UnmappedCompanyName => CompanyName;
-    public string Name { get; set; }
-    public string EmailAddress { get; set; }
+    public string Name { get; set; } = null!;
+    public string EmailAddress { get; set; } = null!;
     public string UnmappedEmailAddress => EmailAddress;
```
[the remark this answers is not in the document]
niemyjski: Already addressed in a prior push — the test models were updated to use \string?\ for properties that can be null in tests (Name, CompanyName, EmailAddress), and generators were adjusted accordingly.

src/Foundatio.Repositories/Models/Aggregations/ObjectValueAggregate.cs
```diff
@@ -9,21 +9,40 @@ namespace Foundatio.Repositories.Models;
...
+            return typed;
 
         if (Value is string stringValue)
-            return serializer.Deserialize<T>(stringValue);
+        {
+            if (serializer is not null)
+                return serializer.Deserialize<T>(stringValue);
+
+            return (T?)Convert.ChangeType(stringValue, typeof(T));
+        }
```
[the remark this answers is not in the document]
niemyjski: The current code (line 27) falls back to Convert.ChangeType for strings without a serializer, which handles simple conversions like string->int. For Nullable<T> targets, the code at line 15 uses Nullable.GetUnderlyingType — wait, that is in TypeHelper. For ObjectValueAggregate specifically, the JsonNode/JsonElement paths already throw InvalidOperationException when serializer is null. The string->Nullable<T> edge case is handled by the caller passing a serializer. This is acceptable behavior.

src/Foundatio.Repositories/JsonPatch/JsonPatcher.cs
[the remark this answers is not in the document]
niemyjski: **Follow-up:** On current `feature/elastic-client`, `Add()` uses `JsonNodeExtensions.UnescapeJsonPointer(parts[^1])` for the property name (JsonPatcher.cs), matching Replace/Remove. RFC 6901 escaping is preserved.

src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticReindexer.cs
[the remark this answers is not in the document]
niemyjski: **Follow-up:** `HandleFailureAsync` already guards with `String.IsNullOrEmpty(failure.Id)`, logs a warning, and returns before `GetAsync` — no `failure.Id!` on the hot path.

src/Foundatio.Repositories.Elasticsearch/Extensions/FindHitExtensions.cs
niemyjski: **Follow-up:** Inline comments in `FindHitExtensions.GetSorts` / `GetFieldValueAsObject` are restored and match the pre–NRT intent (ES client `FieldValue` handling).

src/Foundatio.Repositories.Elasticsearch/Extensions/FindHitExtensions.cs
niemyjski: **Follow-up:** Same as sibling thread — explanatory comments for `IReadOnlyCollection<FieldValue>` and extracting values are kept.

src/Foundatio.Repositories.Elasticsearch/Extensions/FindHitExtensions.cs
niemyjski: **Follow-up:** The “Fallback - return the FieldValue itself” comment remains above the return in `GetFieldValueAsObject`.

src/Foundatio.Repositories.Elasticsearch/Extensions/ElasticIndexExtensions.cs
niemyjski: **Follow-up:** MultiGet `error` callback: early return when `error is null` is now braced; non-null errors are logged with `LogWarning` (see `ToFindHits`).

src/Foundatio.Repositories.Elasticsearch/Extensions/ElasticIndexExtensions.cs
niemyjski: **Follow-up:** MultiGet `result` callback: `result is null` early return is braced; `LogDebug` still runs for not-found documents.

src/Foundatio.Repositories.Elasticsearch/Extensions/ElasticIndexExtensions.cs
niemyjski: **Follow-up:** Date histogram bucket path still includes the “Propagate timezone metadata…” comment; `@timezone` assignment uses an explicit non-null check for the compiler in the lambda.

src/Foundatio.Repositories.Elasticsearch/Extensions/ElasticIndexExtensions.cs
niemyjski: **Follow-up:** Aggregate `Meta` from the ES client is `IReadOnlyDictionary<string, object>`; we keep that shape. For bucket `@timezone`, we now use `hasTimezone && timezoneValue is not null` before assign to avoid `!` and satisfy NRT in the closure.

src/Foundatio.Repositories.Elasticsearch/Extensions/ResolverExtensions.cs
niemyjski: **Follow-up:** `GetResolvedFields` for sorts uses `.Select(...).OfType<SortOptions>().ToList()` — nulls from `ResolveFieldSort` are filtered, no `!` on the projection.

src/Foundatio.Repositories.Elasticsearch/Queries/Builders/FieldConditionsQueryBuilder.cs
niemyjski: **Follow-up:** For `Equals`, `conditionValue` is assigned with `?? throw new QueryValidationException(...)` before `TermQuery`; the value passed to `ToFieldValue` is non-null by construction.

src/Foundatio.Repositories.Elasticsearch/Queries/Builders/IElasticQueryBuilder.cs
niemyjski: **Follow-up:** `GetTimeZoneAsync` uses `await elasticContext.DefaultTimeZone.Invoke().AnyContext()` — configure-await parity with the rest of the library.

src/Foundatio.Repositories.Elasticsearch/Repositories/BulkResult.cs
niemyjski: **Follow-up:** `BulkResult` builds id sets only from items where `e.Id is not null` before `OfType<string>()` — nullable ids from the client are excluded.

src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticReadOnlyRepositoryBase.cs
niemyjski: **Follow-up:** Verified against `main`: `ShouldReturnDocument` returns `false` when `document is null` on both branches; no behavior change vs main.

src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticRepositoryBase.cs
niemyjski: **Follow-up:** After `if (response.Source is not { } source) throw …`, `source` is a non-null `T`; `action?.Invoke(source)` only null-checks the delegate list entries.

src/Foundatio.Repositories.Elasticsearch/ElasticUtility.cs
niemyjski: **Follow-up:** Snapshot enumeration uses `snapshotsResponse.Snapshots ?? []` instead of forcing non-null with `!`.

src/Foundatio.Repositories.Elasticsearch/Utility/FieldValueHelper.cs
niemyjski: **Follow-up (this PR):** `FieldValueHelper` fallback arm now uses `value.ToString() ?? String.Empty` instead of `ToString()!`.

src/Foundatio.Repositories.Elasticsearch/Queries/Builders/FieldConditionsQueryBuilder.cs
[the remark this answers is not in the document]
niemyjski: **Follow-up:** Range branch uses `condition.Value ?? throw new QueryValidationException(...)` only — no duplicate unreachable `ArgumentException` after `ValidateRangeCondition`.

src/Foundatio.Repositories.Elasticsearch/CustomFields/CustomFieldDefinitionRepository.cs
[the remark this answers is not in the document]
niemyjski: **Follow-up:** `InvalidateCacheByQueryAsync` resolves `TenantKey` from field conditions and returns early if missing; cache key uses both entity type and tenant when present.

src/Foundatio.Repositories/Models/Aggregations/ObjectValueAggregate.cs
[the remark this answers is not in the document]
niemyjski: **Follow-up:** `ValueAs<T>` requires a serializer for `JsonNode`/`JsonElement`; otherwise throws `InvalidOperationException`. `Convert.ChangeType` is only for non-JSON `Value` shapes when serializer is omitted.

src/Foundatio.Repositories.Elasticsearch/ElasticUtility.cs
niemyjski: **Follow-up:** Task listing uses `tasksResponse.Nodes?.Values ?? []` to avoid null-forcing `Nodes`.

src/Foundatio.Repositories/Extensions/AggregationsExtensions.cs
niemyjski: **Follow-up:** `aggregations ?? EmptyReadOnly<string, IAggregate>.Dictionary` preserves the previous empty-aggregation semantics for keyed buckets.

src/Foundatio.Repositories/JsonPatch/JsonDiffer.cs
niemyjski: **Follow-up:** `JsonDiffer` uses `String.Equals(add.Path, prevRemove.Path, StringComparison.Ordinal)` when coalescing remove+add.

src/Foundatio.Repositories/Models/Aggregations/ObjectValueAggregate.cs
niemyjski: **Follow-up:** Same as `ValueAs` thread — optional serializer is intentional for primitive paths; JSON-backed values require serializer or throw.

tests/Foundatio.Repositories.Elasticsearch.Tests/Extensions/ElasticsearchExtensions.cs
niemyjski: **Follow-up:** `ElasticsearchExtensions.AssertSingleIndexAlias` uses `Assert.NotNull` on `Aliases`/`Values` before indexing.

tests/Foundatio.Repositories.Elasticsearch.Tests/Repositories/Models/LogEvent.cs
niemyjski: **Follow-up:** Confirmed — `LogEvent.Generate` typing for `stuff`/`id` is already addressed and correct for the factory API.

tests/Foundatio.Repositories.Elasticsearch.Tests/AggregationQueryTests.cs
niemyjski: **Follow-up:** `AggregationQueryTests` round-trip section uses `Assert.NotNull` on aggregations/terms before chaining; no `!` on `Terms` calls in that block.

tests/Foundatio.Repositories.Elasticsearch.Tests/IndexTests.cs
niemyjski: **Follow-up:** `IndexTests` alias section uses `Assert.NotNull` on `Aliases`/`Values` before use.

tests/Foundatio.Repositories.Elasticsearch.Tests/ReindexTests.cs
niemyjski: **Follow-up:** `ReindexTests` alias section uses `Assert.NotNull` on `Aliases`/`Values` before use.

tests/Foundatio.Repositories.Elasticsearch.Tests/RepositoryTests.cs
niemyjski: **Follow-up:** `RepositoryTests` patch test assigns `GetByIdAsync` then `Assert.NotNull(employee)` before assertions — no null-forgiving on the entity.

PR #254, opened 2026-04-10
Fix NRT errors and enable TreatWarningsAsErrors

niemyjski:
## Summary
- Replaced ineffective \<WarningsAsErrors>true</WarningsAsErrors>\ with correct \<TreatWarningsAsErrors>true</TreatWarningsAsErrors>\
- Fixed 160+ nullable reference errors across the Elasticsearch repository implementation

## Key changes
- **FieldConditionQueryExtensions.cs + FieldCondition.cs**: Made value parameters nullable (\object? value\) — the query builder already handles null by rewriting Equals to IsEmpty
- **ElasticRepositoryBase.cs**: Added null guard in bulk save loop, made \ModifiedDocument<T>.Original\ nullable
- **ElasticReadOnlyRepositoryBase.cs**: Guarded nullable query parameter in GetIncludes/GetExcludes
- **LoggerExtensions.cs**: Made elasticResponse parameter nullable (method already null-checks internally)
- **Serialization tests**: Added Assert.NotNull before dereferencing deserialized objects

## Test plan
- [x] \dotnet build Foundatio.All.slnx\ — 0 warnings, 0 errors

PR #263, opened 2026-04-21
fix!: custom field null safety, FindOneAsync Empty, join and Exists guards

niemyjski:
## Summary

Ports null-safety and correctness fixes from the elastic-client investigation to `main` (NEST). Fixes stale custom field data when `ProcessValueAsync` returns null for calculated fields.

> **Breaking Change:** `SetDocumentCustomField` now accepts `object?` and **removes** the field from `Data` when the value is null, rather than storing null. On `v7.17.17` (pre-NRT), calculated fields that returned `{ Value = null }` silently stored null into `Data` because NRT was not enabled. The NRT migration (`#254`) surfaced this mismatch but papered over it with `!` operators. This PR fixes the root cause — null values now result in field removal, which is consistent with how `OnCustomFieldsDocumentsChanging` already skips `customField.Value is null` entries during iteration. Code that relied on `Data.ContainsKey("field")` returning `true` with a null value will now see `false` instead.

### Custom Field Null Handling (Breaking)

- **`SetDocumentCustomField`** — Accepts `object?`; delegates to `RemoveDocumentCustomField` when null
- **`RemoveDocumentCustomField`** — New helper to remove custom fields from `IHaveCustomFields.Data` or `IHaveVirtualCustomFields`
- **Call sites** — Removed `if (result.Value is not null)` guards so null values trigger field removal instead of being silently skipped
- **`OnCustomFieldsDocumentsChanging`** — Clears `Idx` when `CustomFieldDefinitionRepository` is null; logs warning
- **`OnCustomFieldsBeforeQuery`** — Logs warning when `CustomFieldDefinitionRepository` is null

### Null-Safety Guards

- **`ICustomFieldType.ProcessValueAsync`** — `object?` parameter for AlwaysProcess paths where value may be absent
- **`OnCustomFieldsBeforeQuery` / `HandleUnmappedCustomField`** — Early return or `RepositoryException` when definition repo missing
- **Save flows (FullOuterJoin)** — `GetDocumentJoinKey` + safe `ModifiedDocument<T>` construction (no null-forgiving on join sides)
- **`FindOneAsync`** — Returns `FindHit<T>.Empty` instead of null from `FirstOrDefault()`
- **`ExistsAsync(query)`** — `HasIdentity` guard before `_idField` use

### Documentation and Tests

- **XML docs** — Added on all custom field helper methods (`Set`, `Remove`, `Get`, `GetIdx`, `GetCustomFields`, `GetTenantKey`)
- **Tests** — Three new calculated field tests: `CalculatedField_InputsRemoved_ClearsStaleDataValue`, `CalculatedField_PartialInputs_ComputesFromAvailableValues`, `CalculatedField_UpdateTransitionsToNull_RemovesFromDataAndIdx`
- **Test quality** — Proper `Assert.NotNull` patterns (no `!` hacks), `TryGetValue` over `ContainsKey` + indexer

## Testing

- `dotnet build` — clean, 0 warnings
- `Foundatio.Repositories.Tests` — 56 passed
- `Foundatio.Repositories.Elasticsearch.Tests` — 524 passed, 6 skipped (pre-existing)

Review requested: @ejsmith


src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticRepositoryBase.cs
```diff
@@ -1179,19 +1179,27 @@ protected virtual async Task OnCustomFieldsBeforeQuery(object sender, BeforeQuer
...
         var mapping = fieldMapping.ToDictionary(kvp => kvp.Key, kvp => kvp.Value.GetIdxName(), StringComparer.OrdinalIgnoreCase);
 
         args.Options.QueryFieldResolver(mapping.ToHierarchicalFieldResolver("idx."));
     }
 
     protected virtual async Task OnCustomFieldsDocumentsChanging(object sender, DocumentsChangeEventArgs<T> args)
     {
+        var definitionRepo = ElasticIndex.Configuration.CustomFieldDefinitionRepository;
+        if (definitionRepo is null)
+            return;
```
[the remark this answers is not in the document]
niemyjski: Addressed in 27fc17f9. When \definitionRepo is null\, we now iterate documents and clear their \Idx\ dictionaries before returning, preventing stale indexed values.

src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticRepositoryBase.cs
```diff
@@ -1226,11 +1234,16 @@ protected virtual async Task OnCustomFieldsDocumentsChanging(object sender, Docu
...
                     var result = await fieldType.ProcessValueAsync(doc, customField.Value, fieldDefinition);
-                    SetDocumentCustomField(doc, customField.Key, result.Value!);
-                    idx[fieldDefinition.GetIdxName()] = (result.Idx ?? result.Value)!;
+
+                    if (result.Value is not null)
+                        SetDocumentCustomField(doc, customField.Key, result.Value);
+
+                    object? idxValue = result.Idx ?? result.Value;
+                    if (idxValue is not null)
+                        idx[fieldDefinition.GetIdxName()] = idxValue;
```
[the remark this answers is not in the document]
niemyjski: Addressed in 27fc17f9. \SetDocumentCustomField\ now accepts \object?\ and delegates to \RemoveDocumentCustomField\ when the value is null. The \if (result.Value is not null)\ guard has been removed — we always call \SetDocumentCustomField\, which removes the key from \Data\ (or calls \IHaveVirtualCustomFields.RemoveCustomField\) when null. Since \idx.Clear()\ is called at the top of each document's processing, \Idx\ is rebuilt from scratch and null values are naturally excluded.

src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticRepositoryBase.cs
```diff
@@ -1242,12 +1255,17 @@ protected virtual async Task OnCustomFieldsDocumentsChanging(object sender, Docu
...
-                    SetDocumentCustomField(doc, alwaysProcessField.Name, result.Value!);
-                    idx[alwaysProcessField.GetIdxName()] = (result.Idx ?? result.Value)!;
+                    var result = await fieldType.ProcessValueAsync(doc, value, alwaysProcessField);
+
+                    if (result.Value is not null)
+                        SetDocumentCustomField(doc, alwaysProcessField.Name, result.Value);
+
+                    object? alwaysIdxValue = result.Idx ?? result.Value;
+                    if (alwaysIdxValue is not null)
+                        idx[alwaysProcessField.GetIdxName()] = alwaysIdxValue;
```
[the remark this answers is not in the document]
niemyjski: Addressed in 27fc17f9. Same fix as above — the \if (result.Value is not null)\ guard is removed for AlwaysProcess fields too. \SetDocumentCustomField\ always fires and handles null by removing the field. \Idx\ is rebuilt from scratch via \idx.Clear()\ at the top, so null values are never written to it. Three new tests verify calculated fields correctly clear \Data\ and \Idx\ when inputs are removed or produce NaN.

src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticRepositoryBase.cs
```diff
@@ -1179,19 +1179,35 @@ protected virtual async Task OnCustomFieldsBeforeQuery(object sender, BeforeQuer
         if (String.IsNullOrEmpty(tenantKey))
             return;
 
-        var fieldMapping = await ElasticIndex.Configuration.CustomFieldDefinitionRepository!.GetFieldMappingAsync(EntityTypeName, tenantKey);
+        var definitionRepo = ElasticIndex.Configuration.CustomFieldDefinitionRepository;
+        if (definitionRepo is null)
+            return;
```
niemyjski, line 1186: Should we log a warning before returning here?

src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticRepositoryBase.cs
```diff
@@ -1179,19 +1179,35 @@ protected virtual async Task OnCustomFieldsBeforeQuery(object sender, BeforeQuer
...
+        var definitionRepo = ElasticIndex.Configuration.CustomFieldDefinitionRepository;
+        if (definitionRepo is null)
+        {
+            foreach (var doc in args.Documents.Select(d => d.Value))
+            {
+                var idx = GetDocumentIdx(doc);
+                idx?.Clear();
+            }
+
+            return;
```
niemyjski, line 1208: should we be logging a warning here as well?

tests/Foundatio.Repositories.Elasticsearch.Tests/CustomFieldTests.cs
```diff
@@ -663,4 +663,169 @@ await _customFieldDefinitionRepository.AddAsync([
...
+
+        var employee = EmployeeWithCustomFieldsGenerator.Generate(age: 23);
+        employee.CompanyId = "1";
+        employee.Data["Upd1"] = 1;
+        employee.Data["Upd2"] = 2;
+        await _employeeRepository.AddAsync(employee, o => o.ImmediateConsistency());
+
+        var withValue = await _employeeRepository.GetByIdAsync(employee.Id, o => o.Cache(false));
+        Assert.NotNull(withValue);
+        Assert.True(withValue.Data.ContainsKey("UpdCalc"));
```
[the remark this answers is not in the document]
niemyjski: Addressed in d8f47509. Replaced \ContainsKey\ + indexer with \TryGetValue\ for a single dictionary lookup.

src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticRepositoryBase.cs
```diff
@@ -1179,19 +1179,35 @@ protected virtual async Task OnCustomFieldsBeforeQuery(object sender, BeforeQuer
...
     protected virtual async Task OnCustomFieldsDocumentsChanging(object sender, DocumentsChangeEventArgs<T> args)
     {
+        var definitionRepo = ElasticIndex.Configuration.CustomFieldDefinitionRepository;
+        if (definitionRepo is null)
+        {
+            foreach (var doc in args.Documents.Select(d => d.Value))
+            {
+                var idx = GetDocumentIdx(doc);
+                idx?.Clear();
+            }
```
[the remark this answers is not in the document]
niemyjski: Intentionally kept as \oreach\ — the loop body performs side effects (\idx.Clear()\), and using \.Select()\ for side effects is an anti-pattern. The existing \.Select(d => d.Value)\ is just projecting from \ModifiedDocument<T>\ to \T\, which is the idiomatic pattern used throughout this method (see line 1206).

src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticRepositoryBase.cs
```diff
@@ -1262,9 +1286,16 @@ protected virtual async Task OnCustomFieldsDocumentsChanging(object sender, Docu
         if (String.IsNullOrEmpty(tenantKey))
             return null;
 
-        return await ElasticIndex.Configuration.CustomFieldDefinitionRepository!.AddFieldAsync(EntityTypeName, GetDocumentTenantKey(document)!, name, StringFieldType.IndexType);
+        var definitionRepo = ElasticIndex.Configuration.CustomFieldDefinitionRepository;
+        if (definitionRepo is null)
+            throw new RepositoryException("Custom field definition repository is not configured.");
```
niemyjski, line 1296: we throw here but not in others? what's correct?

src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticRepositoryBase.cs
niemyjski: Added in 540bb49e. Logs a warning now: \CustomFieldDefinitionRepository is not configured for index {IndexName}; custom field query resolution skipped\. Kept as a warning + early return rather than a throw since this is an event handler and throwing would break the query pipeline.

src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticRepositoryBase.cs
niemyjski: Added in 540bb49e. Logs a warning: \CustomFieldDefinitionRepository is not configured for index {IndexName}; clearing Idx for all documents\. Same rationale -- warning + graceful degradation since this is a DocumentsChanging event handler.

src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticRepositoryBase.cs
niemyjski: The throw is intentional here -- \HandleUnmappedCustomField\ is called when \AutoCreateCustomFields\ is enabled and a new field needs to be added. Auto-creation fundamentally requires the definition repository, so failing fast with a \RepositoryException\ is correct. The other two locations (\OnCustomFieldsBeforeQuery\ and \OnCustomFieldsDocumentsChanging\) are event handlers where throwing would break the entire query/save pipeline, so they log warnings and degrade gracefully instead.

PR #264, opened 2026-04-22
fix!: throw RepositoryException when custom field helpers are called on unsupported types

niemyjski:
## Summary

- The five protected custom field helper methods (\GetDocumentCustomFields\, \SetDocumentCustomField\, \RemoveDocumentCustomField\, \GetDocumentCustomField\, \GetDocumentIdx\) previously returned \
ull\ or silently no-oped when \T\ did not implement \IHaveCustomFields\ or \IHaveVirtualCustomFields\. They now throw \RepositoryException\ with a clear message.
- Removed dead null checks in \OnCustomFieldsDocumentsChanging\ that were unreachable after the change.
- Added test \CustomFieldHelpers_OnNonCustomFieldType_ThrowRepositoryException\ verifying all five methods throw on a non-custom-field document type.

## Breaking Change

Subclasses calling these protected methods on a document type that does not implement \IHaveCustomFields\ or \IHaveVirtualCustomFields\ will now receive a \RepositoryException\ instead of \
ull\/no-op. Internal callers are unaffected because the custom field event handler is only registered when \HasCustomFields\ is \	rue\.

## Test plan

- [x] New test passes: \CustomFieldHelpers_OnNonCustomFieldType_ThrowRepositoryException\
- [x] All 20 \CustomFieldTests\ pass (no regressions in existing custom field behavior)
- [x] All 56 core repository tests pass
- [x] Build succeeds with zero warnings/errors

### DIFF
--- a/src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticRepositoryBase.cs
+++ b/src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticRepositoryBase.cs
@@ -1202,7 +1202,7 @@ public abstract class ElasticRepositoryBase<T> : ElasticReadOnlyRepositoryBase<T
             foreach (var doc in args.Documents.Select(d => d.Value))
             {
                 var idx = GetDocumentIdx(doc);
-                idx?.Clear();
+                idx.Clear();
             }
 
             return;
@@ -1217,14 +1217,9 @@ public abstract class ElasticRepositoryBase<T> : ElasticReadOnlyRepositoryBase<T
             foreach (var doc in tenant)
             {
                 var idx = GetDocumentIdx(doc);
-                if (idx == null)
-                    continue;
-
                 idx.Clear();
 
                 var customFields = GetDocumentCustomFields(doc);
-                if (customFields == null)
-                    continue;
 
                 foreach (var customField in customFields)
                 {
@@ -1315,13 +1310,13 @@ public abstract class ElasticRepositoryBase<T> : ElasticReadOnlyRepositoryBase<T
     /// Gets the custom fields dictionary from the document (<c>Data</c> for <see cref="IHaveCustomFields"/>,
     /// or the virtual fields collection for <see cref="IHaveVirtualCustomFields"/>).
     /// </summary>
-    protected IDictionary<string, object?>? GetDocumentCustomFields(T document)
+    protected IDictionary<string, object?> GetDocumentCustomFields(T document)
     {
         return document switch
         {
             IHaveCustomFields f => f.Data,
             IHaveVirtualCustomFields v => v.GetCustomFields(),
-            _ => null
+            _ => throw new RepositoryException($"Document type {typeof(T).Name} does not implement IHaveCustomFields or IHaveVirtualCustomFields.")
         };
     }
 
@@ -1345,6 +1340,8 @@ public abstract class ElasticRepositoryBase<T> : ElasticReadOnlyRepositoryBase<T
             case IHaveVirtualCustomFields v:
                 v.SetCustomField(name, value);
                 return;
+            default:
+                throw new RepositoryException($"Document type {typeof(T).Name} does not implement IHaveCustomFields or IHaveVirtualCustomFields.");
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticRepositoryBase.cs
```diff
@@ -1202,7 +1202,7 @@ protected virtual async Task OnCustomFieldsDocumentsChanging(object sender, Docu
             foreach (var doc in args.Documents.Select(d => d.Value))
             {
                 var idx = GetDocumentIdx(doc);
-                idx?.Clear();
+                idx.Clear();
```
niemyjski, line 1205: any reason we still would need this? what cases would docs be null? are there any?

src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticRepositoryBase.cs
```diff
@@ -1217,14 +1217,9 @@ protected virtual async Task OnCustomFieldsDocumentsChanging(object sender, Docu
             foreach (var doc in tenant)
             {
                 var idx = GetDocumentIdx(doc);
```
niemyjski, line 1219: can this return null?

src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticRepositoryBase.cs
```diff
@@ -1217,14 +1217,9 @@ protected virtual async Task OnCustomFieldsDocumentsChanging(object sender, Docu
...
-                if (idx == null)
-                    continue;
-
                 idx.Clear();
 
                 var customFields = GetDocumentCustomFields(doc);
```
niemyjski, line 1222: can this return null?

tests/Foundatio.Repositories.Elasticsearch.Tests/CustomFieldTests.cs
```diff
@@ -827,4 +828,28 @@ await _customFieldDefinitionRepository.AddAsync([
         var noHit = await _employeeRepository.FindAsync(q => q.Company("1").FilterExpression("updcalc:3"), o => o.QueryLogLevel(LogLevel.Information));
         Assert.Empty(noHit.Documents);
     }
+
+    [Fact]
+    public void CustomFieldHelpers_OnNonCustomFieldType_ThrowsRepositoryException()
+    {
```
niemyjski, line 834: missing act,assert,arrange

src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticRepositoryBase.cs
niemyjski: No, this cannot be null. \GetDocumentIdx\ now throws \RepositoryException\ for unmatched types, and both \IHaveCustomFields.Idx\ and \IHaveVirtualCustomFields.Idx\ are declared as non-nullable \IDictionary<string, object>\. The \?.\ was already removed in the first commit — \idx.Clear()\ is correct as-is.

src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticRepositoryBase.cs
niemyjski: No — \GetDocumentIdx\ now throws for unmatched types, and both interface contracts declare \Idx\ as non-nullable. Cannot return null.

src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticRepositoryBase.cs
niemyjski: No — \GetDocumentCustomFields\ now throws for unmatched types. \IHaveData.Data\ and \IHaveVirtualCustomFields.GetCustomFields()\ both return non-nullable dictionaries. Cannot return null.

tests/Foundatio.Repositories.Elasticsearch.Tests/CustomFieldTests.cs
niemyjski: Good catch — added Arrange / Act & Assert comments.

PR #265, opened 2026-04-22
Fix NRT code smells across repository layer

niemyjski:
## Summary

Comprehensive NRT code smell fixes after enabling nullable reference types. Addresses dangerous `!` suppressions, makes interface signatures honest about nullability, and fixes all cascading compiler errors.

**HIGH priority fixes:**
- **ReindexWorkItemHandler**: Remove `!` from `AcquireAsync` (return type is already `Task<ILock?>`), add `ArgumentNullException.ThrowIfNull` for `workItem` instead of `!` suppression
- **MigrationBase._logger**: Initialize to `NullLogger.Instance` instead of `null!` (NRE if parameterless ctor used)
- **MigrationManager.State**: Add `Debug.Assert` to document control-flow invariant instead of `!` suppression
- **IOptions.Get\<T\>/SafeGetOption**: Add `[return: MaybeNull]` so the compiler warns callers who omit `defaultValue` for reference types

**MEDIUM priority fixes:**
- **CommandOptionsDescriptor\<T\>**: Make nullable on all 16 `IRepository`/`IReadOnlyRepository` interface methods and 10 `ElasticRepositoryBase` implementations (without `= null` default to avoid overload ambiguity)
- **CommandOptionsExtensions.As\<T\>** and **SafeGetSet\<T\>**: Make `this` parameter nullable to match existing null-check logic
- **AGENTS.md**: Document NRT conventions for the repository layer

**Cascading fixes from `[return: MaybeNull]`:**
- Nullable return types on ~16 option getter methods (ElasticCommandOptions, SearchAfter, RuntimeFields, FieldIncludes, Expression, Parent)
- `string?` locals in 3 expression query builder classes, ElasticReadOnlyRepositoryBase, and test query
- Null guards in SoftDeletesQueryBuilder (`?.`), ParentQueryBuilder, ChildQueryBuilder

## Test plan

- [ ] CI passes -- `dotnet build` with 0 warnings, 0 errors across all TFMs
- [ ] Existing tests pass with no regressions
- [ ] Verify `CommandOptionsDescriptor<T>?` change is non-breaking (existing lambda callers still compile)
- [ ] Verify `[return: MaybeNull]` doesn't break any callers that pass explicit non-null defaults

### DIFF
--- a/src/Foundatio.Repositories.Elasticsearch/Options/ElasticCommandOptions.cs
+++ b/src/Foundatio.Repositories.Elasticsearch/Options/ElasticCommandOptions.cs
@@ -115,7 +115,7 @@ namespace Foundatio.Repositories.Options
 
         public static Type ParentDocumentType(this ICommandOptions options)
         {
-            return options.SafeGetOption<Type>(ParentDocumentTypeKey, typeof(object))!;
+            return options.SafeGetOption<Type>(ParentDocumentTypeKey, typeof(object));
         }
 
         internal const string QueryFieldResolverKey = "@QueryFieldResolver";
--- a/src/Foundatio.Repositories.Elasticsearch/Options/ElasticCommandOptions.cs
+++ b/src/Foundatio.Repositories.Elasticsearch/Options/ElasticCommandOptions.cs
@@ -68,7 +68,7 @@ namespace Foundatio.Repositories.Options
             return options;
         }
 
-        public static IIndex GetElasticIndex(this ICommandOptions options)
+        public static IIndex? GetElasticIndex(this ICommandOptions options)
         {
             return options.SafeGetOption<IIndex>(ElasticIndexKey);
         }
@@ -97,7 +97,7 @@ namespace Foundatio.Repositories.Options
             return options;
         }
 
-        public static Type DocumentType(this ICommandOptions options)
+        public static Type? DocumentType(this ICommandOptions options)
         {
             return options.SafeGetOption<Type>(DocumentTypeKey);
         }
@@ -115,7 +115,7 @@ namespace Foundatio.Repositories.Options
 
         public static Type ParentDocumentType(this ICommandOptions options)
         {
-            return options.SafeGetOption<Type>(ParentDocumentTypeKey, typeof(object));
+            return options.SafeGetOption<Type>(ParentDocumentTypeKey, typeof(object))!;
         }
 
         internal const string QueryFieldResolverKey = "@QueryFieldResolver";
@@ -125,7 +125,7 @@ namespace Foundatio.Repositories.Options
             return options;
         }
 
-        public static QueryFieldResolver GetQueryFieldResolver(this ICommandOptions options)
+        public static QueryFieldResolver? GetQueryFieldResolver(this ICommandOptions options)
         {
             return options.SafeGetOption<QueryFieldResolver>(QueryFieldResolverKey);
         }
@@ -137,7 +137,7 @@ namespace Foundatio.Repositories.Options
             return options;
         }
 
-        public static IncludeResolver GetIncludeResolver(this ICommandOptions options)
+        public static IncludeResolver? GetIncludeResolver(this ICommandOptions options)
         {
             return options.SafeGetOption<IncludeResolver>(IncludeResolverKey);
         }
--- a/src/Foundatio.Repositories.Elasticsearch/Queries/Builders/ChildQueryBuilder.cs
+++ b/src/Foundatio.Repositories.Elasticsearch/Queries/Builders/ChildQueryBuilder.cs
@@ -57,6 +57,9 @@ namespace Foundatio.Repositories.Elasticsearch.Queries.Builders
                 return;
 
             var index = ctx.Options.GetElasticIndex();
+            if (index is null)
+                return;
+
             foreach (var childQuery in childQueries)
             {
                 var childOptions = ctx.Options.Clone();
--- a/src/Foundatio.Repositories.Elasticsearch/Queries/Builders/ChildQueryBuilder.cs
+++ b/src/Foundatio.Repositories.Elasticsearch/Queries/Builders/ChildQueryBuilder.cs
@@ -56,9 +56,8 @@ namespace Foundatio.Repositories.Elasticsearch.Queries.Builders
             if (childQueries.Count == 0)
                 return;
 
-            var index = ctx.Options.GetElasticIndex();
-            if (index is null)
-                return;
+            var index = ctx.Options.GetElasticIndex()
+                ?? throw new InvalidOperationException("ElasticIndex must be set on options to build child queries.");
 
             foreach (var childQuery in childQueries)
             {
--- a/src/Foundatio.Repositories.Elasticsearch/Queries/Builders/ParentQueryBuilder.cs
+++ b/src/Foundatio.Repositories.Elasticsearch/Queries/Builders/ParentQueryBuilder.cs
@@ -56,7 +56,7 @@ namespace Foundatio.Repositories.Options
             return query.SafeGetOption<(string Relation, string ParentId)>(ParentQueryExtensions.ParentIdKey);
         }
 
-        public static string GetDiscriminator(this IRepositoryQuery query)
+        public static string? GetDiscriminator(this IRepositoryQuery query)
         {
             return query.SafeGetOption<string>(ParentQueryExtensions.DiscriminatorKey);
         }
@@ -75,12 +75,12 @@ namespace Foundatio.Repositories.Elasticsearch.Queries.Builders
             if (!String.IsNullOrEmpty(parentId.Item2))
                 ctx.Filter &= new ParentIdQuery { Id = parentId.ParentId, Type = parentId.Relation };
 
-            string discriminator = ctx.Source.GetDiscriminator();
+            string? discriminator = ctx.Source.GetDiscriminator();
             if (discriminator != null)
                 ctx.Filter &= new TermQuery { Field = "discriminator", Value = discriminator };
 
             var parentQueries = ctx.Source.GetParentQueries();
-            if (parentQueries.Count > 0)
+            if (parentQueries.Count > 0 && index is not null)
             {
                 foreach (var parentQuery in parentQueries)
                 {
--- a/src/Foundatio.Repositories.Elasticsearch/Queries/Builders/ParentQueryBuilder.cs
+++ b/src/Foundatio.Repositories.Elasticsearch/Queries/Builders/ParentQueryBuilder.cs
@@ -69,8 +69,6 @@ namespace Foundatio.Repositories.Elasticsearch.Queries.Builders
     {
         public async Task BuildAsync<T>(QueryBuilderContext<T> ctx) where T : class, new()
         {
-            var index = ctx.Options.GetElasticIndex();
-
             var parentId = ctx.Source.GetParentId();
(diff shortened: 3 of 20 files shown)


src/Foundatio.Repositories.Elasticsearch/Options/ElasticCommandOptions.cs
```diff
@@ -115,7 +115,7 @@ public static T ParentDocumentType<T>(this T options, Type? parentDocumentType)
 
         public static Type ParentDocumentType(this ICommandOptions options)
         {
-            return options.SafeGetOption<Type>(ParentDocumentTypeKey, typeof(object));
+            return options.SafeGetOption<Type>(ParentDocumentTypeKey, typeof(object))!;
```
niemyjski: Feels like this should return Type? and any caller would throw?

src/Foundatio.Repositories.Elasticsearch/Queries/Builders/ChildQueryBuilder.cs
```diff
@@ -57,6 +57,9 @@ public class ChildQueryBuilder : IElasticQueryBuilder
                 return;
 
             var index = ctx.Options.GetElasticIndex();
+            if (index is null)
+                return;
```
niemyjski: should this throw?

src/Foundatio.Repositories.Elasticsearch/Queries/Builders/ParentQueryBuilder.cs
```diff
@@ -75,12 +75,12 @@ public class ParentQueryBuilder : IElasticQueryBuilder
...
                 ctx.Filter &= new ParentIdQuery { Id = parentId.ParentId, Type = parentId.Relation };
 
-            string discriminator = ctx.Source.GetDiscriminator();
+            string? discriminator = ctx.Source.GetDiscriminator();
             if (discriminator != null)
                 ctx.Filter &= new TermQuery { Field = "discriminator", Value = discriminator };
 
             var parentQueries = ctx.Source.GetParentQueries();
-            if (parentQueries.Count > 0)
+            if (parentQueries.Count > 0 && index is not null)
```
niemyjski: should we throw earlier if index is null

src/Foundatio.Repositories.Elasticsearch/Queries/Builders/SoftDeletesQueryBuilder.cs
```diff
@@ -27,7 +27,7 @@ public class SoftDeletesQueryBuilder : IElasticQueryBuilder
             return Task.CompletedTask;
 
         var documentType = ctx.Options.DocumentType();
-        var property = documentType.GetProperty(nameof(ISupportSoftDeletes.IsDeleted));
+        var property = documentType?.GetProperty(nameof(ISupportSoftDeletes.IsDeleted));
```
niemyjski: should this throw as a fallback if type cannot be resolved

src/Foundatio.Repositories/Options/IOptions.cs
```diff
@@ -113,15 +116,15 @@ public static ICollection<T> SafeGetCollection<T>(this IOptions? options, string
...
+        return options.Values.Get(name, new List<T>())!;
     }
 
     public static TOptions AddCollectionOptionValue<TOptions, TValue>(this TOptions options, string name, TValue value) where TOptions : IOptions
     {
         if (options == null)
             throw new ArgumentNullException(nameof(options));
 
-        var setOption = options.SafeGetOption(name, new List<TValue>());
+        var setOption = options.SafeGetOption(name, new List<TValue>())!;
```
niemyjski: why do we do ! here? is it due to fallback?

src/Foundatio.Repositories/Options/IOptions.cs
```diff
@@ -133,27 +136,27 @@ public static TOptions AddCollectionOptionValue<TOptions, TValue>(this TOptions
...
         setOption.AddRange(values);
         options.Values.Set(name, setOption);
 
         return options;
     }
 
-    public static ISet<T> SafeGetSet<T>(this IOptions options, string name)
+    public static ISet<T> SafeGetSet<T>(this IOptions? options, string name)
```
niemyjski, line 148: why do we do ! here? is it due to fallback?

src/Foundatio.Repositories/Options/IOptions.cs
```diff
@@ -165,7 +168,7 @@ public static T AddSetOptionValue<T>(this T options, string name, IEnumerable<st
         if (options == null)
             throw new ArgumentNullException(nameof(options));
 
-        var setOption = options.SafeGetOption(name, new HashSet<string>(StringComparer.OrdinalIgnoreCase));
+        var setOption = options.SafeGetOption(name, new HashSet<string>(StringComparer.OrdinalIgnoreCase))!;
```
niemyjski: why do we do ! here? is it due to fallback?

src/Foundatio.Repositories.Elasticsearch/Options/ElasticCommandOptions.cs
niemyjski: `ParentDocumentType` uses `typeof(object)` as the sentinel default, so with `[return: NotNullIfNotNull]` the compiler now knows this never returns null -- the `!` is removed. The `typeof(object)` sentinel pattern is itself questionable (it could be `Type?` with `null` meaning "not configured") but that is a larger refactor beyond NRT scope.

src/Foundatio.Repositories.Elasticsearch/Queries/Builders/ChildQueryBuilder.cs
niemyjski: Changed to throw `InvalidOperationException` instead of silently returning. If you have child queries configured but no index set, that is a configuration bug that should fail loudly.

src/Foundatio.Repositories.Elasticsearch/Queries/Builders/ParentQueryBuilder.cs
niemyjski: Changed to throw `InvalidOperationException` when parent queries exist but index is null. The throw only fires inside the `parentQueries.Count > 0` guard, so the method still returns cleanly when there are no parent queries.

src/Foundatio.Repositories.Elasticsearch/Queries/Builders/SoftDeletesQueryBuilder.cs
niemyjski: Changed to throw `InvalidOperationException` when `SupportsSoftDeletes` is true but `DocumentType()` returns null. If soft deletes is enabled, the type must be known to resolve the `IsDeleted` field name correctly.

src/Foundatio.Repositories/Options/IOptions.cs
niemyjski: Added `[return: NotNullIfNotNull(nameof(defaultValue))]` to both `IOptionsDictionary.Get<T>` and `SafeGetOption<T>`. This tells the compiler that when a non-null default is passed, the return is guaranteed non-null. All 6 `!` suppressions in this file are now removed -- the compiler handles it correctly.

src/Foundatio.Repositories/Options/IOptions.cs
niemyjski: Same fix -- `[return: NotNullIfNotNull(nameof(defaultValue))]` on the method declarations eliminates the need for `!` here. Removed.

src/Foundatio.Repositories/Options/IOptions.cs
niemyjski: Same as above -- `NotNullIfNotNull` attribute handles this now. All `!` suppressions removed from callers that pass non-null defaults.

src/Foundatio.Repositories/Options/IOptions.cs
```diff
@@ -113,15 +116,15 @@ public static ICollection<T> SafeGetCollection<T>(this IOptions? options, string
...
 
     public static TOptions AddCollectionOptionValue<TOptions, TValue>(this TOptions options, string name, TValue value) where TOptions : IOptions
     {
         if (options == null)
             throw new ArgumentNullException(nameof(options));
 
-        var setOption = options.SafeGetOption(name, new List<TValue>());
+        var setOption = options.SafeGetOption(name, new List<TValue>())!;
         setOption.Add(value);
         options.Values.Set(name, setOption);
```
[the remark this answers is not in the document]
niemyjski: Done -- added \[return: NotNullIfNotNull(nameof(defaultValue))]\ to both \IOptionsDictionary.Get<T>\ and \SafeGetOption<T>\. The \!\ suppressions on \AddCollectionOptionValue\ and all similar helpers have been removed. The compiler now correctly infers non-nullability when a non-null default is passed.

src/Foundatio.Repositories/Options/IOptions.cs
```diff
@@ -133,27 +136,27 @@ public static TOptions AddCollectionOptionValue<TOptions, TValue>(this TOptions
...
 
-    public static ISet<T> SafeGetSet<T>(this IOptions options, string name)
+    public static ISet<T> SafeGetSet<T>(this IOptions? options, string name)
     {
         if (options == null)
             return new HashSet<T>();
 
-        return options.Values.Get(name, new HashSet<T>());
+        return options.Values.Get(name, new HashSet<T>())!;
     }
```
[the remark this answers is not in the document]
niemyjski: Same fix applied -- \[return: NotNullIfNotNull(nameof(defaultValue))]\ on the method declarations removes the need for \!\ on \SafeGetSet\ and \AddSetOptionValue\ as well.

src/Foundatio.Repositories/Options/TimeProviderOptions.cs
```diff
@@ -20,7 +20,7 @@ public static class ReadTimeProviderOptionsExtensions
     {
         public static TimeProvider GetTimeProvider(this IOptions options)
         {
-            return options.SafeGetOption(SetTimeProviderOptionsExtensions.TimeProviderKey, TimeProvider.System);
+            return options.SafeGetOption(SetTimeProviderOptionsExtensions.TimeProviderKey, TimeProvider.System)!;
```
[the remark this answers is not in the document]
niemyjski: Exactly right. Added \[return: NotNullIfNotNull(nameof(defaultValue))]\ to \SafeGetOption\ and \IOptionsDictionary.Get\. The \!\ on \GetTimeProvider()\ is now removed -- the compiler understands \TimeProvider.System\ as non-null default guarantees a non-null return.

src/Foundatio.Repositories/Options/IOptions.cs
```diff
@@ -113,15 +116,15 @@ public static ICollection<T> SafeGetCollection<T>(this IOptions? options, string
         if (options == null)
             return new List<T>();
 
-        return options.Values.Get(name, new List<T>());
+        return options.Values.Get(name, new List<T>())!;
     }
```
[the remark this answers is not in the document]
niemyjski: Agreed. Added \[return: NotNullIfNotNull(nameof(defaultValue))]\ to both \IOptionsDictionary.Get<T>\ and \SafeGetOption<T>\. All \!\ suppressions on callers that pass non-null defaults (SafeGetCollection, AddCollectionOptionValue, SafeGetSet, AddSetOptionValue, GetTimeProvider, ParentDocumentType) have been removed.

PR #272, opened 2026-05-11
Fix concurrency and correctness issues in reindex and bulk operations

niemyjski:
## Summary

Improves correctness, operational safety, and notification precision across Elasticsearch reindexing, bulk delete, and patch operations.

### Reindex Safety

- **DailyIndex.ReindexAsync distributed lock**: Added distributed lock acquisition matching the pattern in `VersionedIndex.ReindexAsync`. Both use `AcquireAsync(lockKey, TimeSpan.FromMinutes(20), TimeSpan.FromMinutes(30))` which auto-throws `LockAcquisitionTimeoutException`. Prevents concurrent migration jobs from silently overwriting newer data.

- **Reindex second-pass typed sampling**: `GetSampleDocumentIdAsync` now returns a typed `SampleIdResult` (Found/Empty/Failed) instead of a nullable string. This disambiguates an empty source index from a transient search failure:
  - **Empty**: logs info and skips the second pass.
  - **Found + ObjectId-compatible IDs**: uses ObjectId-based range queries as fallback for the second pass.
  - **Found + non-ObjectId IDs**: logs Warning about the limitation (not Critical, since it is not process-fatal).
  - **Failed**: logs Warning with the search error and attempts the ObjectId-based second pass anyway, since skipping guarantees no catch-up while the range reindex may still succeed.

### Bulk Operations

- **Bulk delete retry with exponential backoff**: Added retry logic for transient 429/503 errors, matching the existing `IndexDocumentsBulkAsync` pattern. Accumulates both fatal and conflict IDs across all attempts, then checks `finalResult.HasErrors` to throw — ensuring HTTP 409 conflicts are never silently swallowed.

- **PatchAllAsync per-batch notifications**: Notifications are now sent per-batch inline (bounded by page size) instead of accumulating an unbounded list of all modified IDs and fanning out at the end. Retried IDs (from version conflict retry) are excluded from batch notifications since `PatchAsync` already notifies for them.

### Documentation

- **PatchAllAsync XML docs**: Documented Conflicts.Proceed behavior so callers understand version conflict semantics.
- **Updated docs**: index-management.md, jobs.md, crud-operations.md, patch-operations.md, message-bus.md, and SKILL.md updated for new behaviors.

## Test plan

- [x] Build passes: `dotnet build Foundatio.Repositories.slnx`
- [x] All tests pass (CI green on both net8 and net10)
- [x] DailyIndex concurrent reindex test: `ReindexTimeSeriesIndex_ConcurrentCalls_OnlyOneReindexes`
- [x] VersionedIndex concurrent reindex tests: `ReindexAsync_ConcurrentCalls_OnlyOneCompletes`, `ReindexAsync_LockPreventsHandlerAndDirectConcurrency`
- [x] Bulk delete test: `BulkRemoveAsync_MultipleDocuments_RemovesAllAndNotifies`
- [x] PatchAllAsync no-op tests: `PatchAllAsync_ScriptNoOp_DoesNotSendNotification`, `PatchAllAsync_ActionPatchNoOp_DoesNotSendNotification`
- [x] PatchAllAsync partial notification: `PatchAllAsync_PartialModification_NotifiesOnlyModifiedDocs`
- [x] Per-ID notification test: `PatchAllAsync_FilterQuery_SendsPerIdNotification`


src/Foundatio.Repositories.Elasticsearch/Configuration/DailyIndex.cs
```diff
@@ -228,6 +229,16 @@ public override async Task ReindexAsync(Func<int, string?, Task>? progressCallba
         if (currentVersion < 0 || currentVersion >= Version)
             return;
 
+        string lockKey = ElasticReindexer.GetLockName(Name);
+        using var lockCts = new CancellationTokenSource(TimeSpan.FromMinutes(30));
+        await using var reindexLock = await Configuration.LockProvider.TryAcquireAsync(lockKey, TimeSpan.FromMinutes(20), cancellationToken: lockCts.Token).AnyContext();
```
niemyjski: Is it an issue with all these indexes that the cancellation token is 10 minutes longer than the acquire timeout? or what is the cancellation token, reindexing could take longer than 30 minutes so this seems really bad. should this be the disposed token instead of the lockCts, can you look at all implementations.

src/Foundatio.Repositories.Elasticsearch/Configuration/DailyIndex.cs
```diff
@@ -228,6 +229,16 @@ public override async Task ReindexAsync(Func<int, string?, Task>? progressCallba
...
+        if (reindexLock is null)
+            throw new InvalidOperationException($"Unable to acquire reindex lock for '{Name}' after 30 minutes.");
```
niemyjski: it would be better to use the AcquireAsync which automatically throws then throw this invalid operation exception. look at the other implementations and fix as well.

src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticReindexer.cs
```diff
@@ -124,6 +124,24 @@ public async Task ReindexAsync(ReindexWorkItem workItem, Func<int, string?, Task
...
             await progressCallbackAsync(97, $"Total: {secondPassResult.Total:N0} Completed: {secondPassResult.Completed:N0}").AnyContext();
         }
+        else
+        {
+            var sampleId = await GetSampleDocumentIdAsync(workItem.OldIndex).AnyContext();
+            if (sampleId is not null && ObjectId.TryParse(sampleId, out var unused))
+            {
+                _logger.LogInformation("Reindex {OldIndex} -> {NewIndex}: Using ObjectId-based second pass (no TimestampField).", workItem.OldIndex, workItem.NewIndex);
+                secondPassResult = await InternalReindexAsync(workItem, progressCallbackAsync, 92, 96, startTime).AnyContext();
+                if (!secondPassResult.Succeeded) return;
```
niemyjski: We need logging? and return on separate line.

src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticReindexer.cs
```diff
@@ -124,6 +124,24 @@ public async Task ReindexAsync(ReindexWorkItem workItem, Func<int, string?, Task
 
             await progressCallbackAsync(97, $"Total: {secondPassResult.Total:N0} Completed: {secondPassResult.Completed:N0}").AnyContext();
         }
+        else
+        {
+            var sampleId = await GetSampleDocumentIdAsync(workItem.OldIndex).AnyContext();
+            if (sampleId is not null && ObjectId.TryParse(sampleId, out var unused))
```
niemyjski: do other implementations have this check for looking at object id and second pass? Feels like this could be a helper method???

src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticRepositoryBase.cs
```diff
@@ -613,32 +613,55 @@ public virtual async Task RemoveAsync(IEnumerable<T> documents, ICommandOptions?
         }
         else
         {
-            var response = await _client.BulkAsync(bulk =>
+            IReadOnlyCollection<T> docsToDelete = docs;
+            var allSuccessfulDocs = new List<T>();
+            var allFatalIds = new HashSet<string>();
```
niemyjski: should these have the capacity set?

src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticRepositoryBase.cs
```diff
@@ -613,32 +613,55 @@ public virtual async Task RemoveAsync(IEnumerable<T> documents, ICommandOptions?
...
+                    : docsToDelete.Where(d => result.SuccessfulIds.Contains(d.Id)).ToList();
+                allSuccessfulDocs.AddRange(successDocs);
+                allFatalIds.UnionWith(result.FatalIds);
 
-            if (result.HasErrors)
+                if (!result.HasRetryableErrors || attempt >= 3)
+                    break;
+
+                docsToDelete = docsToDelete.Where(d => result.RetryableIds.Contains(d.Id)).ToList();
+                var delay = TimeSpan.FromSeconds(Math.Pow(2, attempt));
```
niemyjski, line 656: don't we have a helper for this?

src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticRepositoryBase.cs
```diff
@@ -613,32 +613,55 @@ public virtual async Task RemoveAsync(IEnumerable<T> documents, ICommandOptions?
...
+
+                if (DisposedCancellationToken.IsCancellationRequested)
+                    break;
+            }
+
+            if (allSuccessfulDocs.Count > 0)
+                await OnDocumentsRemovedAsync(allSuccessfulDocs, options).AnyContext();
+
+            if (allFatalIds.Count > 0 || result.HasRetryableErrors)
                 ThrowForBulkErrors(result, operationLabel: "removing");
```
niemyjski: would this throw know about the other fatal ids? Are we logging all of this state out?

src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticRepositoryBase.cs
```diff
@@ -848,13 +881,15 @@ await BatchProcessAsync(query, async results =>
 
                 if (modifiedDocuments.Count > 0 || retriedIds.Count > 0)
                 {
+                    var batchIds = modifiedDocuments.Select(h => h.Id!).Concat(retriedIds).ToList();
```
niemyjski, line 891: do we have to have all these allocations?

src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticRepositoryBase.cs
```diff
@@ -984,10 +1019,10 @@ await BatchProcessAsync(query, async results =>
                         .Where(h => result.SuccessfulIds.Contains(h.Id!) && !result.NoopIds.Contains(h.Id!))
                         .Select(h => h.Id!).ToList();
                     modifiedInBatch += updatedIds.Count;
+                    allModifiedIds.AddRange(updatedIds);
 
                     if (IsCacheEnabled && updatedIds.Count > 0)
                     {
-                        // TODO: Invalidate by documents instead of IDs to support custom cache invalidation overrides.
```
niemyjski, line 990: I really don't like us removing code comments like this, restore all code comments unless the todo was actually implemented.

src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticRepositoryBase.cs
```diff
@@ -1011,15 +1046,15 @@ await BatchProcessAsync(query, async results =>
 
         if (affectedRecords > 0)
         {
-            // When called via PatchAsync(Ids), query has IDs and this duplicates per-batch invalidation
-            // (harmless/idempotent). When called directly with a filter query, GetIds() is empty (no-op).
             if (IsCacheEnabled)
                 await InvalidateCacheByQueryAsync(query.As<T>()).AnyContext();
 
-            // Empty list: batch callbacks handle per-document cache invalidation for ActionPatch/JsonPatch.
-            // This fires the DocumentsChanged event so subscribers know documents were modified.
```
niemyjski, line 1020: restore

tests/Foundatio.Repositories.Elasticsearch.Tests/RepositoryTests.cs
```diff
@@ -2643,10 +2663,13 @@ await _employeeRepository.PatchAllAsync(
             new ScriptPatch("ctx._source.name = 'Changed';"),
             o => o.ImmediateConsistency());
 
-        // Assert
-        using var cts = new CancellationTokenSource(TimeSpan.FromSeconds(2));
+        // Assert — per-ID notifications sent for each modified document
```
niemyjski, line 2666: what's the overhead of per id? Guess it's nice but what if it's in the thousands? How many tasks are we awaiting?

niemyjski:
All review feedback has been addressed in commit 23ab85b9:

**Lock pattern (DailyIndex/VersionedIndex):** Switched to AcquireAsync extension method which auto-throws LockAcquisitionTimeoutException. Removed manual null check, InvalidOperationException, and unnecessary CancellationTokenSource.

**ElasticReindexer:** Null sampleId (empty source index) now handled gracefully with info log and early skip of second pass. Return placed on separate line.

**Bulk delete aggregation:** allSuccessfulDocs now has capacity set. ThrowForBulkErrors receives aggregated fatal IDs via record with expression. SafeDelay pattern is consistent with IndexDocumentsBulkAsync.

**PatchAllAsync unbounded list:** Replaced allModifiedIds accumulation with per-batch inline notifications. Each batch sends its own SendNotificationsAsync call immediately, avoiding unbounded memory growth. A notifiedPerBatch flag ensures the final fallback only sends type-level notification when no batch path was taken.

**Restored comments:** Restored the TODO and explanatory comments that were inadvertently removed.

src/Foundatio.Repositories.Elasticsearch/Configuration/DailyIndex.cs
niemyjski: Fixed in 23ab85b9. Removed the CancellationTokenSource entirely. Now using `AcquireAsync(lockKey, TimeSpan.FromMinutes(20), TimeSpan.FromMinutes(30))` extension — lock expiration is 20 min (renewed in callback), acquire timeout is 30 min. Auto-throws `LockAcquisitionTimeoutException`.

src/Foundatio.Repositories.Elasticsearch/Configuration/DailyIndex.cs
niemyjski: Done in 23ab85b9. Both DailyIndex and VersionedIndex now use `AcquireAsync` extension which auto-throws.

src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticReindexer.cs
niemyjski: Fixed in 23ab85b9. Added info logging for empty index case and separated return onto its own line.

src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticReindexer.cs
niemyjski: This is the only place this check exists — there is no other implementation that does the ObjectId fallback. Kept it inline since it is a single-use flow. If we add more callers we can extract.

src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticRepositoryBase.cs
niemyjski: Set capacity on `allSuccessfulDocs` in 23ab85b9: `new List<T>(docs.Count)`.

src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticRepositoryBase.cs
niemyjski: This uses the same `SafeDelay` + `DisposedCancellationToken` pattern as `IndexDocumentsBulkAsync` (line ~1767). No separate helper exists — this is the standard pattern in the codebase.

src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticRepositoryBase.cs
niemyjski: Fixed in 23ab85b9. Now using `result with { FatalIds = allFatalIds }` so ThrowForBulkErrors receives the aggregated fatal IDs from all attempts. The BulkResponse itself is logged per-attempt via `_logger.LogRequest(response)`.

src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticRepositoryBase.cs
```diff
@@ -613,32 +613,55 @@ public virtual async Task RemoveAsync(IEnumerable<T> documents, ICommandOptions?
...
 
-            var result = BulkResult.From(response);
+                result = BulkResult.From(response);
 
-            var successDocs = result.IsSuccess ? docs : docs.Where(d => result.SuccessfulIds.Contains(d.Id)).ToList();
-            if (successDocs.Count > 0)
-                await OnDocumentsRemovedAsync(successDocs, options).AnyContext();
+                var successDocs = result.IsSuccess
+                    ? docsToDelete.ToList()
+                    : docsToDelete.Where(d => result.SuccessfulIds.Contains(d.Id)).ToList();
```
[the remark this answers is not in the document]
niemyjski: Fixed in 23ab85b9. `allFatalIds` is accumulated across all attempts and passed to `ThrowForBulkErrors` via `result with { FatalIds = allFatalIds }`.

src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticRepositoryBase.cs
```diff
@@ -613,32 +613,55 @@ public virtual async Task RemoveAsync(IEnumerable<T> documents, ICommandOptions?
...
+            for (int attempt = 0; attempt < 4; attempt++)
             {
-                bulk.Refresh(options.GetRefreshMode(DefaultConsistency));
-                foreach (var doc in docs)
-                    bulk.Delete<T>(d =>
-                    {
-                        d.Id(doc.Id).Index(ElasticIndex.GetIndex(doc));
+                var response = await _client.BulkAsync(bulk =>
+                {
+                    bulk.Refresh(options.GetRefreshMode(DefaultConsistency));
```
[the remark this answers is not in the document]
niemyjski: Acknowledged — the retry behavior follows the same pattern as `IndexDocumentsBulkAsync`. Integration test `BulkRemoveAsync_MultipleDocuments_RemovesAllAndNotifies` covers the happy path; simulating 429/503 would require a mock transport layer.

src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticRepositoryBase.cs
```diff
@@ -613,32 +613,55 @@ public virtual async Task RemoveAsync(IEnumerable<T> documents, ICommandOptions?
...
-            if (successDocs.Count > 0)
-                await OnDocumentsRemovedAsync(successDocs, options).AnyContext();
+                var successDocs = result.IsSuccess
+                    ? docsToDelete.ToList()
+                    : docsToDelete.Where(d => result.SuccessfulIds.Contains(d.Id)).ToList();
+                allSuccessfulDocs.AddRange(successDocs);
+                allFatalIds.UnionWith(result.FatalIds);
 
-            if (result.HasErrors)
+                if (!result.HasRetryableErrors || attempt >= 3)
```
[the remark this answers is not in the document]
niemyjski: Fixed in 23ab85b9 with `result with { FatalIds = allFatalIds }` — aggregates across all retry attempts.

src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticRepositoryBase.cs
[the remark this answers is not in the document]
niemyjski: Fixed in 23ab85b9. ThrowForBulkErrors now receives the aggregated `allFatalIds` via record `with` expression.

src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticRepositoryBase.cs
```diff
@@ -693,6 +723,7 @@ public virtual async Task<long> PatchAllAsync(IRepositoryQuery query, IPatchOper
             operation = ApplyDateTracking(partialPatchOp);
 
         long affectedRecords = 0;
+        var allModifiedIds = new List<string>();
```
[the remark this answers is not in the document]
niemyjski: Fixed in 23ab85b9. Replaced the unbounded `allModifiedIds` list with per-batch inline notifications. Each batch sends `SendNotificationsAsync` immediately — no accumulation, no end-of-method fan-out.

src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticReindexer.cs
```diff
@@ -124,6 +124,24 @@ public async Task ReindexAsync(ReindexWorkItem workItem, Func<int, string?, Task
...
+                if (!secondPassResult.Succeeded) return;
+
+                await progressCallbackAsync(97, $"Total: {secondPassResult.Total:N0} Completed: {secondPassResult.Completed:N0}").AnyContext();
+            }
+            else
+            {
+                _logger.LogCritical(
+                    "Reindex {OldIndex} -> {NewIndex}: No TimestampField and IDs are not ObjectIds. Cannot perform second-pass catch-up. Documents written during reindex may be lost. Consider adding IHaveDates to your model or using ObjectId-format IDs.",
+                    workItem.OldIndex, workItem.NewIndex);
+            }
```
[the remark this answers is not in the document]
niemyjski: Fixed in 23ab85b9. Null sampleId (empty index) is now handled separately with an info log before the ObjectId check. The LogCritical path only fires when a non-null sampleId fails ObjectId parsing.

src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticRepositoryBase.cs
niemyjski: These per-batch allocations are bounded by batch size (the search page size, typically 1000). This matches the existing pattern in the JsonPatch path above. The big win was removing the unbounded `allModifiedIds` accumulation.

src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticRepositoryBase.cs
niemyjski: Restored in 23ab85b9.

src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticRepositoryBase.cs
niemyjski: Restored in 23ab85b9.

tests/Foundatio.Repositories.Elasticsearch.Tests/RepositoryTests.cs
niemyjski: Good question. With the latest fix, notifications are now sent per-batch (not accumulated across all batches). Each batch sends one `Task.WhenAll` over its batch IDs (bounded by search page size). No unbounded fan-out at the end anymore.

src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticRepositoryBase.cs
```diff
@@ -613,33 +613,58 @@ public virtual async Task RemoveAsync(IEnumerable<T> documents, ICommandOptions?
...
+                    break;
+            }
+
+            if (allSuccessfulDocs.Count > 0)
+                await OnDocumentsRemovedAsync(allSuccessfulDocs, options).AnyContext();
+
+            if (allFatalIds.Count > 0 || result.HasRetryableErrors)
+            {
+                ThrowForBulkErrors(result with { FatalIds = allFatalIds }, operationLabel: "removing");
+            }
```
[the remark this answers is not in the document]
niemyjski: Fixed in 9122e8a6. Added `result.HasTransportError` to the throw condition: `if (allFatalIds.Count > 0 || result.HasRetryableErrors || result.HasTransportError)`.

tests/Foundatio.Repositories.Elasticsearch.Tests/RepositoryTests.cs
```diff
@@ -2757,4 +2780,85 @@ await _employeeRepository.PatchAllAsync(
...
+
+        // Act — noop script that doesn't modify anything
+        await _employeeRepository.PatchAllAsync(
+            q => q,
+            new ScriptPatch("ctx.op = 'none';"),
+            o => o.ImmediateConsistency());
+
+        // Assert — no documents were modified, so no notifications should be sent
+        await Task.Delay(100, TestCancellationToken);
+        Assert.Equal(0, _messageBus.MessagesSent);
```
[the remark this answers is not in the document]
niemyjski: Fixed in 9122e8a6. Removed the `Task.Delay` — since notifications are awaited inline, the assertion is reliable without a delay.

tests/Foundatio.Repositories.Elasticsearch.Tests/RepositoryTests.cs
```diff
@@ -2757,4 +2780,85 @@ await _employeeRepository.PatchAllAsync(
...
+
+        // Act — action that returns false (no modification)
+        await _employeeRepository.PatchAllAsync(
+            q => q,
+            new ActionPatch<Employee>(e => false),
+            o => o.ImmediateConsistency());
+
+        // Assert — no documents were modified, so no notifications should be sent
+        await Task.Delay(100, TestCancellationToken);
+        Assert.Equal(0, _messageBus.MessagesSent);
```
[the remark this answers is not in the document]
niemyjski: Fixed in 9122e8a6. Removed the `Task.Delay`.

src/Foundatio.Repositories.Elasticsearch/Configuration/VersionedIndex.cs
```diff
@@ -253,8 +253,6 @@ public override async Task ReindexAsync(Func<int, string?, Task>? progressCallba
 
         string lockKey = ElasticReindexer.GetLockName(Name);
         await using var reindexLock = await Configuration.LockProvider.AcquireAsync(lockKey, TimeSpan.FromMinutes(20), TimeSpan.FromMinutes(30)).AnyContext();
-        if (reindexLock is null)
-            throw new InvalidOperationException($"Unable to acquire reindex lock for '{Name}' after 30 minutes.");
```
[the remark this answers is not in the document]
niemyjski: Not applicable. We are using the `AcquireAsync` **extension method** (from `Foundatio.Lock.ILockProviderExtensions`) which throws `LockAcquisitionTimeoutException` — it never returns null. The non-nullable return type `Task<ILock>` (not `Task<ILock?>`) confirms this.

src/Foundatio.Repositories.Elasticsearch/Configuration/DailyIndex.cs
```diff
@@ -228,6 +229,13 @@ public override async Task ReindexAsync(Func<int, string?, Task>? progressCallba
         if (currentVersion < 0 || currentVersion >= Version)
             return;
 
+        string lockKey = ElasticReindexer.GetLockName(Name);
+        await using var reindexLock = await Configuration.LockProvider.AcquireAsync(lockKey, TimeSpan.FromMinutes(20), TimeSpan.FromMinutes(30)).AnyContext();
+
```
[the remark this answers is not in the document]
niemyjski: Not applicable. Same as above — the `AcquireAsync` extension method (3-arg overload with `TimeSpan? acquireTimeout`) throws `LockAcquisitionTimeoutException` on failure and never returns null.

src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticRepositoryBase.cs
```diff
@@ -755,13 +788,17 @@ await BatchProcessAsync(query, async results =>
 
                 if (modifiedDocuments.Count > 0 || retriedIds.Count > 0)
                 {
+                    var batchIds = successfulIds.Concat(retriedIds).ToList();
+                    notifiedPerBatch = true;
```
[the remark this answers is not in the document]
niemyjski: Good catch. Fixed in 9122e8a6. `SendNotificationsAsync` now only sends for `successfulIds` (bulk-successful), not `retriedIds`. The retried IDs are already notified by `PatchAsync` inside `HandleBulkPatchErrorsAsync`.

src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticRepositoryBase.cs
```diff
@@ -848,13 +885,17 @@ await BatchProcessAsync(query, async results =>
 
                 if (modifiedDocuments.Count > 0 || retriedIds.Count > 0)
                 {
+                    var batchIds = modifiedDocuments.Select(h => h.Id!).Concat(retriedIds).ToList();
+                    notifiedPerBatch = true;
+
                     if (IsCacheEnabled)
                         await InvalidateCacheAsync(modifiedDocuments.Select(h => h.Document!)).AnyContext();
 
+                    await SendNotificationsAsync(ChangeType.Saved, (IReadOnlyCollection<string>)batchIds, options).AnyContext();
```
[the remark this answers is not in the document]
niemyjski: Fixed in 9122e8a6. Same approach — `SendNotificationsAsync` now sends only for `bulkSuccessIds`, not the concatenated `batchIds` that includes `retriedIds`.

src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticRepositoryBase.cs
```diff
@@ -886,12 +888,15 @@ await BatchProcessAsync(query, async results =>
...
+                    var bulkSuccessIds = modifiedDocuments.Select(h => h.Id!).ToList();
                     notifiedPerBatch = true;
 
                     if (IsCacheEnabled)
                         await InvalidateCacheAsync(modifiedDocuments.Select(h => h.Document!)).AnyContext();
 
-                    await SendNotificationsAsync(ChangeType.Saved, (IReadOnlyCollection<string>)batchIds, options).AnyContext();
+                    // Notify only for bulk-successful IDs; retriedIds already notified by PatchAsync in HandleBulkPatchErrorsAsync.
+                    if (bulkSuccessIds.Count > 0)
+                        await SendNotificationsAsync(ChangeType.Saved, (IReadOnlyCollection<string>)bulkSuccessIds, options).AnyContext();
```
niemyjski: it really sucks we do all this casting, can we make this better.

niemyjski: Agreed. The casts were unnecessary — `List<string>` already resolves to the `IReadOnlyCollection<string>` overload without ambiguity. Removed all three in f3b0a5a5.

src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticRepositoryBase.cs
```diff
@@ -613,33 +613,58 @@ public virtual async Task RemoveAsync(IEnumerable<T> documents, ICommandOptions?
...
+                    break;
+            }
+
+            if (allSuccessfulDocs.Count > 0)
+                await OnDocumentsRemovedAsync(allSuccessfulDocs, options).AnyContext();
+
+            if (allFatalIds.Count > 0 || result.HasRetryableErrors || result.HasTransportError)
+            {
+                ThrowForBulkErrors(result with { FatalIds = allFatalIds }, operationLabel: "removing");
+            }
```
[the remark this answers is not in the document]
niemyjski: Intentional. For bulk deletes, a 409 conflict means the document was concurrently modified — it will be caught on the next delete pass or is being handled by another operation. This matches the existing `IndexDocumentsBulkAsync` pattern where conflicts are tracked separately and not treated as fatal for deletes.

src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticReindexer.cs
```diff
@@ -451,6 +474,21 @@ private QueryContainer CreateRangeQuery(QueryContainerDescriptor<object> descrip
...
+    private async Task<string?> GetSampleDocumentIdAsync(string index)
+    {
+        var response = await _client.SearchAsync<IDictionary<string, object>>(d => d
+            .Index(index)
+            .Source(s => s.ExcludeAll())
+            .Size(1)
+        ).AnyContext();
+
+        _logger.LogRequest(response);
+        if (!response.IsValid || !response.Hits.Any())
```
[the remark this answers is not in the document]
niemyjski: Acknowledged. In practice, a null return from a search failure vs empty index has the same consequence: skip the second pass. The first pass already copied all existing docs. The only risk is docs written *during* reindex, which is the documented limitation. Adding search-failure detection would add complexity for a very rare transient edge case — the reindex would need to be re-run regardless.

src/Foundatio.Repositories.Elasticsearch/Configuration/DailyIndex.cs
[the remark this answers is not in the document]
niemyjski: The description was from an earlier iteration. We switched to `AcquireAsync` (which auto-throws) per reviewer feedback. The code is correct; the description is stale context from iteration history.

src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticRepositoryBase.cs
niemyjski: Fixed in f3b0a5a5. Removed all `(IReadOnlyCollection<string>)` casts — `List<string>` already resolves to the correct overload without ambiguity.

src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticRepositoryBase.cs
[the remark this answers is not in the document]
niemyjski: Intentional for deletes. A 409 on a delete means the document was concurrently modified — it will be caught on the next delete pass. This matches existing behavior and is acceptable for idempotent delete semantics.

src/Foundatio.Repositories.Elasticsearch/Repositories/ElasticReindexer.cs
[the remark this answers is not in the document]
niemyjski: Acknowledged. Both cases (empty index, search failure) have the same outcome: skip the second pass. The first pass already copied all existing docs. A transient search failure here means the reindex should be re-run anyway. The risk (docs written during reindex being missed) is inherent and documented in the LogCritical path.

src/Foundatio.Repositories.Elasticsearch/Configuration/DailyIndex.cs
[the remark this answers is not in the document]
niemyjski: PR description has been updated in 2303c65c to reflect the current `AcquireAsync` approach.

